PR #10, opened 2022-04-01
Code generation only prints with `owlkettleDebug` defined.

### LINKED ISSUE #5: Macros `echo result.repr`
I was going to make a PR to fix this but given that you have it in multiple positions I feel like you desire this behaviour. If so can it be moved behind a  `owlkettleDebug` flag to not pollute the terminal on project build?

beef331:
This will make it easier to see compiler issues and allow easy debugging if needed.
Closes #5


can-lehmann:
Generally owlkettle uses `snake_case` instead of `camelCase` for identifiers (including compiler flags). It therefore should be called `owlkettle_debug` (with a lower case "d").

can-lehmann:
Looks good!

PR #11, opened 2022-04-03
added activate to Entry

beef331:


can-lehmann:
Looks good!

can-lehmann:
For future reference: The documentation for the widgets is actually automatically generated from the `owlkettle/widgets.nim` file. So there is no need to manually update it to reflect your changes. Since I noticed that is not super obvious how to do this, I added [a brief tutorial](https://github.com/can-lehmann/owlkettle/blob/main/CONTRIBUTING.md#documentation) to the CONTRIBUTING.md file.

PR #12, opened 2022-04-03
Minor improvments to the macros

beef331:
There are some more massive changes that can be made to this like for instance:
```nim
     var cond = new_tree(nnkIfStmt, [
        new_tree(nnkElifBranch, [
          new_dot_expr(widget, "has_" & field.name),
          new_stmt_list(new_assignment(
            new_dot_expr(state, field.name),
            new_dot_expr(widget, "val_" & field.name)
          ))
        ])
      ])
```
Into the following:
```nim
     var cond = genAst(widget, state, field = ident(field.name), hasField = ident("has" & field.name), valField = ident("val" * field.name):
       if widget.hasField:
         state.field = widget.valField
```
As those changes are are rather extreme I'll just mention it and if you want me to make a PR for it can make a series of small ones to make it easier to check(I'll try my darnedest to write the silly snake way :smile:) .

### DIFF
--- a/owlkettle/common.nim
+++ b/owlkettle/common.nim
@@ -24,13 +24,9 @@
 
 import std/[macros, strutils]
 
-proc is_name*(node: NimNode): bool =
-  if node.is_nil:
-    return false
-  result = node.kind == nnkIdent or node.kind == nnkSym
+proc is_name*(node: NimNode): bool = node.kind in {nnkIdent, nnkSym}
 
-proc is_name*(node: NimNode, name: string): bool =
-  result = node.is_name and nim_ident_normalize(node.str_val) == nim_ident_normalize(name)
+proc is_name*(node: NimNode, name: string): bool = node.is_name and node.eq_ident(name)
 
 proc unwrap_name*(node: NimNode): NimNode =
   result = node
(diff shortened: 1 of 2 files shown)


can-lehmann:
Looks good so far. I was a bit concerned over removing the nil check in the `is_name` procedure, but apparently `node.kind` returns `nnkNilLit` if the given node is `nil` (which is imo a somewhat questionable design decision on Nim's part).

I did not know about the `genasts` module, but from what I can see, it seems to make generating code much nicer. I would welcome a PR for using it instead of generating them manually. Maybe do just the `guidsl` module first, so that in case I see any issues with the use of the `gen_ast` macro, you don't have to make too many changes. I would also recommend just using `gen_ast` for larger ASTs (like the one in your example).

One issue I see right now, is that the lines where `gen_ast` is called get very long, maybe something like this is nicer (also `silly_snake_way` :smile:):

```nim
var cond = gen_ast(widget, state,
                   field = ident(field.name),
                   has_field = ident("has" & field.name),
                   val_field = ident("val" & field.name)):
  if widget.has_field:
    state.field = widget.val_field
```

PR #19, opened 2022-10-23
Fix build with Nim devel

planetis-m:
nil can no longer match to distinct pointer type.


planetis-m:
Nim changelog entry: https://github.com/nim-lang/Nim/blob/devel/changelog.md "nil literals can no longer..."

can-lehmann:
Looks good. Thank you for contributing!

PR #22, opened 2022-11-12
Fix compilation for nimPreviewSlimSystem

planetis-m:
This flag moves procs out of system such as $ for floats and assert. It will be default in nim v2. Made sure that every example compiles.


can-lehmann:
Looks good!

planetis-m:
Actually my bad, it's not part of v2: https://github.com/nim-lang/Nim/pull/20761#issuecomment-1311348838

PR #27, opened 2023-02-06
ADD developer tooling documentation

### LINKED ISSUE #26: Add mention of gtk4-icon-browser for figuring out available icons to docs
First up, thanks for making the library! I'm new to GUI toolkits like GTK/QT (originally webdev) and I've been playing around with this for a couple hours and had quite a bit of fun with how quickly and easily everything fit together, particularly how easy it was to split Widgets into their own files and have them share data!

I was fiddling around with the todo-list tutorial and I noticed it using a couple icons here or there. That got me wondering what the available icons were. After 15 minutes of googling and some luck I stumbled over a mention of gtk4-icon-browser. A bit of more googling later on how to install that and I had a chance to see what icons there are. It would be nice if we could shorten that link from "Oh, what icons can I have?" to finding that list for newcomers such as myself.

Imo it makes sense to  mention in the docs what tooling there is out there that one likely needs, such as gtk4-icon-browser to see a list of available icons. If there are more things out there that one should know about I'd love to of course also know about those ^^ 

I'd be happy to write a 3-4 line paragraph for this, but I'm not sure where this should be mentioned. A separate "recommended-tools.md" ? As part of "widgets.md" at the bottom?

PhilippMDoerner:
This provides an overview over tools that simplify development or provide necessary information, such as a list of available icons.

Closes https://github.com/can-lehmann/owlkettle/issues/26

Note:

I have no clue if what I wrote about colors is actually accurate. Custom stylesheets are the only way known to me so far that actually allow you to change colors, that's why I wrote that there. If there are other ways, please state so and I'll add them as examples for usecases to the Colors section.

### DIFF
--- a/docs/recommended_tools.md
+++ b/docs/recommended_tools.md
@@ -20,7 +20,15 @@ For an overview over the color-palette used by gnome, you can install [ColorPale
 
 Generally when developing with custom colors you may want to ensure that the contrast difference between e.g. text and background is sufficient to be easy to read. This is particularly important for accessibility! 
 
-To check if the contrast between 2 pixels on your desktop is large enough, you can use [Contrast](https://flathub.org/apps/details/org.gnome.design.Contrast).
+To check if the contrast between 2 pixels on your desktop is large enough, you can install [Contrast](https://flathub.org/apps/details/org.gnome.design.Contrast).
+
+## GTK Documentation
+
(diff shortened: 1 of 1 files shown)


can-lehmann:
Custom style sheets are indeed the main application of the Color Palette App for owlkettle. I also used it while developing apps that use `DrawingArea` in the past, but I think that is a more niche use case (probably not worth mentioning in the document).

PR #29, opened 2023-02-08
Fix for nim devel

planetis-m:
Similar to #19


can-lehmann:
Thank you! Maybe I should add nim devel to the CI so I don't keep breaking it by accident :smile:

PR #31, opened 2023-02-12
Feature/#28 add developer tool docs

PhilippMDoerner:
This is an in-between PR.
I'll likely make further PR's under this ticket-number, but this way you can have several smaller/medium sized PRs that improve things incrementally rather than one bombastic ones.

Basically all of this PR is just adding a bunch of doc-comments everywhere.
Either to explain what a value stands for, repeat an explanation of a value that has already been written but for a different Widget (vAlign and hAlign have that issue), listing the possible enum-values to make it quicker to find the available values etc.

### DIFF
--- a/docs/internals.md
+++ b/docs/internals.md
@@ -622,7 +622,57 @@ brew(gui(App()))
 Note that there is no `update` or `build` hook defined for the `text` field. If we had defined those, they would need to include sections that call each individual `property` hook like their default implementations would. 
 
 ### **Read Hook**
-# TODO: Write this
+The `read` hook is a custom hook solely used by widgets that are renderables and deal with user input. They may need a hook at times that sends back the current state within a `Widget` whenever the user changes said state through interaction. This hook is then responsible for propagating the changes done by the user from the state of `Widget` to `WidgetState`.
+
+Let's look at a minimal example of the `ColorChooserDialog`:
+
+```nim
+
+import owlkettle
+import owlkettle/[gtk, widgetdef]
+
+# The custom widget
+renderable MyColorChooserDialog of BuiltinDialog:
+  color: tuple[r, g, b, a: float] = (0.0, 0.0, 0.0, 1.0)
+  
+  hooks:
+    beforeBuild: ## Necessary for renderable to instantiate Widget in general
+      state.internalWidget = gtk_color_chooser_dialog_new(
+        widget.valTitle.cstring,
+        nil.GtkWidget
+      )
+  
+  hooks color:
+    read: ## Will execute after userinput was provided, propagates value to to `WidgetState`
+      var color: GdkRgba
+      gtk_color_chooser_get_rgba(state.internalWidget, color.addr)
+      echo "ReadHook - old state color: ", state.color.repr
+      state.color = (color.r.float, color.g.float, color.b.float, color.a.float)
+      echo "ReadHook - new state color: ", state.color.repr
+
+## The App
+viewable App:
+  discard
+
+method view(app: AppState): Widget =
+  result = gui:
+    Window:
+      HeaderBar {.addTitlebar.}:
+        Button {.addLeft.}:
+          text = "Open"
+          style = {ButtonSuggested}
+          proc clicked() =
+            let (res, state) = app.open: gui:
+              MyColorChooserDialog()
+
+brew(gui(App()))
+```
+
+The `read` hook will execute after a color was chosen and confirmed.
+Only then will it update the state with the chosen color. How to extract the data of the user-interaction will depend on the gtk-widget being wrapped.
+
+Note that during the choosing of the color and when updating `WidgetState` via the `read` hook, no `update` hook is being executed.
+
 
 
 ```mermaid
--- a/docs/internals.md
+++ b/docs/internals.md
@@ -241,7 +241,7 @@ The available hooks are:
   - (W) connectEvents   : Only relevant for renderables. Executed every time a callback is supposed to be attached to the underlying GTK widget. It defines how to do so.
   - (W) disconnectEvents: Only relevant for renderables. Executed every time a callback is supposed to be removed from the underlying GTK widget. It defines how to do so.
   - (WF) update         : Executed every time `WidgetState` is updated by `Widget`.
-  - (F) property        : Executed every time the hook-field changed its value in both the update and build phases.
+  - (F) property        : Executed every time the hook-field changed its value during the update or build phases.
   - (F) read            : Used in `Dialog`. Executes every time the state of the underlying GTK-Widget changes.
 
 W: Can act as hook for Widgets
@@ -253,10 +253,10 @@ With the exception of `read`, all hooks also have implicit access to a variable 
 
 Generally the `build`, `property` and `update` hooks are likely to have the highest utility for you. Consult their individual sections for more information.
 
-### Build Hook
+### **Build Hook**
 The `build` hook runs once just before any values are assigned to the `WidgetState`.
 
-#### For Widgets
+#### **For Widgets**
 `build` hooks on widgets should be used when additional logic is necessary that sets multiple fields on `WidgetState` during widget instantiation. Note that such fields should not have assigned default values, as they will be overwritten when default values get applied after the build phase.
 
 Example: A Widget may need to load data from elsewhere, via a file or HTTP request for one field, and a second field must be inferred from a value of the first field.
@@ -348,7 +348,7 @@ Given that the purpose of `beforeBuild` is to handle instantiating renderables a
 
 For more info on the purpose of `beforeBuild` and `afterBuild` hooks, consult their respective sections in this file.
 
-#### For Fields
+#### **For Fields**
 Owlkettle provides default `build` hooks for every field. They are useful if you need simple custom behaviour, such as modifying the input slightly before initially assigning it to a field. It is their responsibility to transfer data from `Widget` to their field in `WidgetState` during the build-phase.
 
 `build` hooks on fields should be used when additional logic is necessary that sets this single field on `WidgetState`. 
@@ -388,7 +388,7 @@ brew(gui(App()))
 
 Note that this hook is not run during updates, so any changes here may be lost if an update overwrites them. Look at the `update` hook if you need that behaviour as well.
 
-### Before-Build Hook
+### **Before-Build Hook**
 The `beforeBuild` hook runs once before the build-hook and thus also before any values are assigned to the `WidgetState`.
 
 Their main usecase is renderables, where they are used to instantiate the GTK-Widget and assign it to `internalWidget` on `WidgetState`.
@@ -426,7 +426,7 @@ But what if we want to have a parent widget decide the text to render once and t
 
 That leads us to what `afterBuild` hooks are...
 
-### After-Build Hook
+### **After-Build Hook**
 The `afterBuild`-hook runs once after initial values (default-values, values passed in by other widgets during instantiation) have been assigned to the `WidgetState`.
 
 They are useful if any processing on the initial data that is passed in must happen. Example are validating data, inferring data from passed in data, or fetching other data based on what was passed in. In renderables they are also useful to update the GTK widget once with data from the initial `WidgetState`.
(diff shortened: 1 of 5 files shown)


can-lehmann:
The doc comments look good so far.
I am not sure if they are correctly transferred to the `widgets.md` file though. Did you try running `make -C docs` in the owlkettle directory?

PhilippMDoerner:
Done

can-lehmann:
The introduction does not seem to render properly, because it is indented:

image

PhilippMDoerner:
> The introduction does not seem to render properly, because it is indented:
> 
> image

Done

owlkettle/widgets.nim
```diff
@@ -28,15 +28,24 @@ when defined(nimPreviewSlimSystem):
 import gtk, widgetdef, cairo, widgetutils
 
 when defined(owlkettleDocs) and isMainModule:
-  echo "# Widgets\n\n"
+  echo """
+# Widgets
+
+## Custom Widgets
+When writing your application or custom widgets you will typically be writing a `viewable` (see [internals.md](https://github.com/can-lehmann/owlkettle/blob/main/docs/internals.md) for more information). All `viewables` have the following available to them by default:
+  - BaseWidget-fields and setters
```
can-lehmann, line 36: This is not correct. Viewables just inherit from `Widget`, not BaseWidget. (Even though the name `BaseWidget` would suggest that)

owlkettle/widgets.nim
```diff
@@ -1320,7 +1346,9 @@ renderable MenuButton of BaseWidget:
   child: Widget
   popover: Widget
   
-  style: set[ButtonStyle]
+  style: set[ButtonStyle] ## Applies special styling to the button. 
+  ## May be one of `ButtonSuggested`, `ButtonDestructive`, `ButtonFlat`, `ButtonPill` or `ButtonCircular`.
+  ## Consult the ## [GTK4 documentation](https://developer.gnome.org/hig/patterns/controls/buttons.html?highlight=button#button-styles) for guidance on what to use.
```
can-lehmann, line 1351: Multiline comments are currently not transferred to `widgets.md`. Just convert this to a single line.

owlkettle/widgets.nim
```diff
@@ -243,13 +252,16 @@ renderable Box of BaseWidget:
               vAlign: AlignFill.}:
     ## Adds a child to the Box.
     ## When expand is true, the child grows to fill up the remaining space in the Box.
-    ## The hAlign and vAlign properties allow you to set the alignment of the child
-    ## within its allocated area.
+    ## The hAlign and vAlign properties allow you to set the horizontal and vertical 
+    ## alignment of the child within its allocated area. They may be one of `AlignFill`, 
+    ## `AlignStart`, `AlignEnd` or `AlignCenter`.
+    ## 
+    ## Note: **Any** widgets contained in a Box-Widget get access the `expand` adder, to control their behaviour inside of the Box!
```
can-lehmann, line 259: I would remove this note. I think how adders work should be described somewhere in general. (Also: `expand` would be called a property here, since it is a parameter to the `Box.add` adder. So e.g. `{.name: "value".}` → Property; `{.name.}` → Adder; `{.name, prop: value.}` → Adder with property)

owlkettle/widgets.nim
```diff
@@ -2101,7 +2129,9 @@ proc toGtk(resp: DialogResponse): cint =
 renderable DialogButton:
   text: string
   response: DialogResponse
-  style: set[ButtonStyle]
+  style: set[ButtonStyle] ## Applies special styling to the button. 
+  ## May be one of `ButtonSuggested`, `ButtonDestructive`, `ButtonFlat`, `ButtonPill` or `ButtonCircular`.
+  ## Consult the ## [GTK4 documentation](https://developer.gnome.org/hig/patterns/controls/buttons.html?highlight=button#button-styles) for guidance on what to use.
```
can-lehmann, line 2134: Multiline comment (see above)

owlkettle/widgets.nim
PhilippMDoerner: I'll try to come up with a decent text-addition to internals.md and remove the note. I'll need to find some proper time to hammer that one out so I'll have to postpone that to "later", but I'll see what I can do!

owlkettle/widgets.nim
PhilippMDoerner: In that case I think it makes sense to scratch that idea for a "Custom Widgets" section in the widgets-docs from me I think.
Maybe that makes more sense as part of the internals.md file as part of explaining what features a Viewable has etc.

PhilippMDoerner: At this point the internals.md section is getting a lot of extra content pumped into it that provides this kind of information and this section has been removed from widgets.nim. I'll thusly mark this as resolved.

owlkettle/widgets.nim
PhilippMDoerner: Done

docs/internals.md
```diff
@@ -1,10 +1,433 @@
...
+renderable MyRenderable:
+  text: string
+    
+  hooks:
+    beforeBuild:
+      echo state.repr
+      state.internalWidget = gtk_label_new("")
+
+    afterBuild:
+      gtk_label_set_text(state.internalWidget, state.text.cstring)
```
can-lehmann, line 381: This should probably use a property hook. You could also set a static text in `gtk_label_new` to simplify the example even more.

owlkettle/widgets.nim
```diff
@@ -321,17 +326,25 @@ type LabelStyle* = enum
...
+  ## Supports rendering [Pango Markup](https://docs.gtk.org/Pango/pango_markup.html#pango-markup) 
+  ## if `useMarkup` is enabled.
+  text: string ## The text of the Label to render
   xAlign: float = 0.5
   yAlign: float = 0.5
-  ellipsize: EllipsizeMode
-  wrap: bool = false
-  useMarkup: bool = false
+  ellipsize: EllipsizeMode ## Determines whether to ellipsise the text in case space is insufficient to render all of it.
+  ## May be one of `EllipsizeNone`, `EllipsizeStart`, `EllipsizeMiddle` or `EllipsizeEnd`
```
can-lehmann, line 343: Multiline comment

owlkettle/widgets.nim
```diff
@@ -417,7 +430,9 @@ type ButtonStyle* = enum
   ButtonCircular = "circular"
 
 renderable Button of BaseWidget:
-  style: set[ButtonStyle]
+  style: set[ButtonStyle] ## Applies special styling to the button. 
+  ## May be one of `ButtonSuggested`, `ButtonDestructive`, `ButtonFlat`, `ButtonPill` or `ButtonCircular`.
+  ## Consult the [GTK4 documentation](https://developer.gnome.org/hig/patterns/controls/buttons.html?highlight=button#button-styles) for guidance on what to use.
```
can-lehmann, line 435: Multiline comment

docs/internals.md
```diff
@@ -1,10 +1,433 @@
...
+Their main usecase is renderables, where they are used to instantiate the GTK-Widget and assign it to `internalWidget` on `WidgetState`.
+
+Here a simple code-example:
+
+```nim
+import owlkettle
+import owlkettle/gtk
+import std/json
+
+type Config = object
```
can-lehmann, line 369: Config is unused.

docs/internals.md
```diff
@@ -1,10 +1,433 @@
...
+
+Here a simple code-example:
+
+```nim
+import std/json
+
+type Config = object
+  name: string
+
+viewable MyViewable:
```
can-lehmann, line 272: Maybe make this the `App` to simplify the example.

owlkettle/widgets.nim
PhilippMDoerner: Done

owlkettle/widgets.nim
PhilippMDoerner: Done

docs/internals.md
PhilippMDoerner: In what sense unused?
It is used in line 273 as the type for a field on `MyViewable`, in line 277 it is assigned to in the `build` hook and used in line 282 in the `view` method to get the name.

PhilippMDoerner: Done

docs/internals.md
PhilippMDoerner: Done. Also added the `json` file so that the example is more complete.

docs/internals.md
PhilippMDoerner: I update this section to ineed just do `gtk_label_new`, because that reduces the noise as much as possible.

As for `afterBuild`, I reused that example in the `afterBuild` section, since it provides a better example for what `afterBuild` can be useful for than what stood there previously (see line 390 and following). It's still with the same flaw (So if you update the parent this child will not change), but imo that makes sense if you want a renderable and only set its values once. It provides a nice guarantee that these "immutable" widgets in a sense are not going to do strange stuff on you regardless of how the AppState changes, so I can see usecases for it.

I explicitly pointed out in that section that one of the short-comings of this approach is that it doesn't update and to look at the property-section for an approach where it does (which I still have to write).

PhilippMDoerner:
I've finished writing up docs for all hooks and their usecases. There's definitely still some improvements to be made, but I'm comfortable saying that this is at least a finished first draft. At this point I'd focus on fixing any comments you provide and fix any mistakes I've made while writing this.

docs/internals.md
```diff
@@ -1,10 +1,679 @@
...
+  result = gui:
+    Window:
+      MyViewable(text = "Example")
+
+brew(gui(App()))
+```
+
+Note that this hook is not run during updates, so any changes here may be lost if an update overwrites them. Look at the `update` hook if you need that behaviour, or `property` hook if you need both.
+
+### **Before-Build Hook**
```
can-lehmann, line 391: `BeforeBuild` for consistency with the other headings.

docs/internals.md
```diff
@@ -1,10 +1,679 @@
...
+
+brew(gui(App()))
+```
+
+We set the label-text to render directly via the `gtk_label_new` proc.
+But what if we want to have a parent widget decide the text to render once and then never update it again?  
+
+That leads us to what `afterBuild` hooks are...
+
+### **After-Build Hook**
```
can-lehmann, line 429: `AfterBuild` for consistency with the other headings.

docs/internals.md
```diff
@@ -1,10 +1,679 @@
...
+
+The available hooks are:
+  - (W) beforeBuild     : Executed only once before the `WidgetState` is created.
+  - (WF) build           : Executed once after `WidgetState` is instantiated from `Widget`. Default values have not yet been applied and will overwrite any values set within this hook.
+  - (W) afterBuild      : Executed only once after the `WidgetState` was created. Is executed *after* all default values have been applied.
+  - (W) connectEvents   : Only relevant for renderables. Executed every time a callback is supposed to be attached to the underlying GTK widget. It defines how to do so.
+  - (W) disconnectEvents: Only relevant for renderables. Executed every time a callback is supposed to be removed from the underlying GTK widget. It defines how to do so.
+  - (WF) update         : Executed every time `WidgetState` is updated by `Widget`.
+  - (F) property        : Executed every time the hook-field changed its value during the update or build phases.
+  - (F) read            : Used in `Dialog`. Executes every time the state of the underlying GTK-Widget changes.
```
can-lehmann, line 245: Maybe make this a markdown table with the columns `Widget/Field` (TODO: better name?), `Name`, `Description`

can-lehmann:
Overall this already looks really good! Its great that we will finally have good documentation for some of the more advanced features. Two general comments:

1. It look like `widgets.md` needs to be updated to reflect some of the changes to `widgets.nim`
2. `internals.md` still contains the old diagrams & text. I think that the diagrams would fit in well with your introduction text on the difference between viewables and renderables, so they could be integrated there.

docs/internals.md
```diff
@@ -1,10 +1,679 @@
...
+
+## The custom widget
+viewable MyViewable:
+  text: string
+      
+  hooks text:
+    update:
+      echo "Received via Widget:    ", widget.valText
+      state.text = widget.valText & " update hook addition"
+      echo "Applied to WidgetState: ", state.text
```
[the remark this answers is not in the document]
PhilippMDoerner: Done. Though question, should this also be done for widget-update-hooks? In the example there I currently don't check for `hasText` either, should I update this?

docs/internals.md
PhilippMDoerner: Done.
I also added the information whether a hook is useful for a viewable, a renderable or both, since certain hooks (beforeBuild, (dis)connectEvents) make little sense on a viewable.

Feel free to tell me whether that should be reverted or not

PhilippMDoerner:
> Overall this already looks really good! Its great that we will finally have good documentation for some of the more advanced features. Two general comments:
> 
>     1. It look like `widgets.md` needs to be updated to reflect some of the changes to `widgets.nim`
> 
>     2. `internals.md` still contains the old diagrams & text at the bottom. I think that the diagrams would fit in well with your introduction text on the difference between viewables and renderables, so they could be integrated there. The old text is probably redundant and should be removed.

Both are done for now.
I think I'll have to give the Widget-Basic another update at one point simply because I think I could make easier to understand with a few images or more concise text, but that'll need me to be hit by a bus of inspiration first on how to do so I think.

PR #34, opened 2023-02-22
Feature/nimibook branch2

PhilippMDoerner:
DO NOT MERGE YET, ONLY REVIEW FOR NOW!

This PR introduces nimibook docs to owlkettle.
In doing so it:

-  replaces "internals.md" with a nimibook-friendly version where each section is split into its own nimibook file.
The key files here are `nbook.nim` and `nimib.toml`, the first defining the order of entries and which entries are being used, the toml defining [nimibook configuration](https://pietroppeter.github.io/nimibook/configuration.html).
- introduces 2 nimble tasks, one to generate the book CLI tool, the second to generate that CLI and generate the docs from there. It also puts the generated folders and files in gitignore.
- Due to mermaid-js not being supported, I replaced them with screenshots of the corresponding graphs in "internals"
- it introduces a github workflow that generates the docs every time main is pushed or PR'd to, similar as the tests.
- It adds links to the book to the README... well it will eventually once the workflow ran on the owlkettle repo once so it deploys the book.
That is also the reason this shouldn't be merged for now, as for testing purposes it currently also reacts to pushes to this testbranch `feature/nimiboo-branch2`. 

For reference, [this](https://philippmdoerner.github.io/owlkettle/README) is the current example for looking at. Once that is approved I'll remove my branch from the `.github/docs.yml` and then it is mergeable.


PhilippMDoerner:
I have no concept as to why it worked all of a sudden in the docs.

The tests went haywire because of rate limitations:
> Error: 403 - {"message":"API rate limit exceeded for 172.176.163.144. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}

As for deployment of the docs, that requires you to set up a repository secret called "TOKEN" , whose value stems from a token generated in Settings -> Developer Settings -> Personal Access Tokens -> (Either fine grained token or Tokens(classic)).

So the general stuff appears to work there.

PhilippMDoerner:
In my eyes this PR is finished, barring implementing any feedback should there be issues with the current implementation.
What would need to be done is:
1) Add a Token as secret to the repository, which I can't do as I'm not the repository owner
2) Rerun the pipelines as they currently fail mostly because some API-call-limit was reached (well, the deploy-docs pipeline fails because the repo has no Token among its secrets)

can-lehmann:
Great! I will take a look at it later. Is the version at https://philippmdoerner.github.io/owlkettle/README current?

PhilippMDoerner:
Aye, that link is current. Once the review is done (and will remain so as long as my branch is included in the docs.yml, which it currently is. That's one of the changes I can only remove once we're done with reviewing these changes ^^)

PhilippMDoerner:
Ahhh shnuckums, the relative links are proving to be my default in the examples/README.md.
I'll try to get those fixed.

PhilippMDoerner:
Okay, fixed the link issue, **now** it should solely need adding a token-secret to the repository so that the deploy-docs workflow can succeed and then review.

README.md
```diff
@@ -31,12 +31,12 @@ brew(gui(App()))
 
 The code above will result in the following application:
 
-<img alt="Counter Application" src="docs/assets/introduction.png" width="228px">
+<img alt="Counter Application" src="../docs/assets/introduction.png" width="228px">
```
can-lehmann, line 34: This file still needs to work as the README of the GitHub Repository, currently the images are not displayed correctly there.

README.md
```diff
@@ -31,12 +31,12 @@ brew(gui(App()))
...
 
 Owlkettle also supports building libadwaita apps.
 To enable libadwaita, import `owlkettle/adw` and change the last line to `adw.brew(gui(App()))`.
 
-<img alt="Counter Application using Adwaita Stylesheet" src="docs/assets/introduction_adwaita.png" width="322px">
+<img alt="Counter Application using Adwaita Stylesheet" src="../docs/assets/introduction_adwaita.png" width="322px">
```
can-lehmann, line 39: This file still needs to work as the README of the GitHub Repository, currently the images are not displayed correctly there.

README.md
```diff
@@ -49,13 +49,11 @@ $ nimble install owlkettle
...
-You can find a reference of all widgets in [docs/widgets.md](docs/widgets.md).
+[Click here]() for reference documentation of all widgets, including libadwaita widgets. 
 
-A tutorial for building a simple Todo application can be found [here](docs/tutorial.md).
+[Click here]() for a tutorial for building a simple Todo application.
 
-Additional examples can be found in the [examples](examples) folder.
-
-A reference of all supported libadwaita widgets can be found in [docs/widgets_adwaita.md](docs/widgets_adwaita.md).
+[Click here]() for additional examples Additional examples can be found in the [examples](examples) folder.
```
can-lehmann, line 56: I would not change the text in this section. Currently the links also don't work neither on the website nor in the GitHub README.

book/internals/hooks/after_build_hook.nim
```diff
@@ -0,0 +1,46 @@
...
+
+  viewable App:
+    discard
+
+  method view(app: AppState): Widget =
+    result = gui:
+      Window:
+        MyRenderable(text = "Defined by App")
+
+  # brew(gui(App())) # Uncomment to execute app
```
can-lehmann, line 39: Should not be commented out in the docs. Maybe we can add a flag like `-d:owlkettleDoNotBrew` to just directly `return` from `brew`

book/internals/adders/adders_with_properties.nim
```diff
@@ -0,0 +1,70 @@
...
+First we need to modify our adder, telling it that there will be additional parameters. 
+
+```nim
+...
+viewable CustomBox:
+  myChildren: Table[string, Widget] # The child-widget field
+
+  adder add {.key: none(string).}: 
+...
+```
```
can-lehmann, line 17: The closing ` ``` ` are shown on the website for some reason, maybe a nimib bug?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,71 @@
+name: github pages
+
+on:
+  push:
+    branches:
+      - main
+      - feature/nimibook-branch2
```
can-lehmann, line 7: Are the docs also deployed when a pull request is created? If so, that would be a giant security issue.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,71 @@
...
+  pull_request:
```
can-lehmann, line 8: Are the docs also deployed when a pull request is created? If so, that would be a giant security issue.

can-lehmann:
I made some minor corrections of the `internals.md` text after merging, those should be applied here as well: https://github.com/can-lehmann/owlkettle/commit/c211d4c0f9f2d6f8bea5da322ae45ea0ad2e9294

README.md
PhilippMDoerner: I may not be seeing what you are seeing there.
The 2 places that need to work are:
https://github.com/PhilippMDoerner/owlkettle/tree/feature/nimibook-branch2
and
https://philippmdoerner.github.io/owlkettle/examples/README.html

Which render to me as:
image
and
image

What am I missing?

README.md
PhilippMDoerner: They can't yet work because what I'd want to link to there would be pages on the nimibook docs once they're deployed on the owlkettle gihub pages. Those don't exist yet since it hasn't been deployed yet, so basically those can only be added in later, or I can start anticipating what the link will look like once its deployed, but then it still won't work yet before the PR is merged.

.github/workflows/docs.yml
PhilippMDoerner: Ah, good catch! Yeah that was a terrible idea, fixed

README.md
can-lehmann: This file is the main README of the repository, not the examples/README.md file.

README.md
can-lehmann: Ah, ok.

book/internals/adders/adders_with_properties.nim
PhilippMDoerner: Yeah that's a nimibook bug. As is, that the rest of the nbText block will be treated as part of the code block.
I workarounded that for now, by leaving the backticks out. That renders properly for now, and by closing the nbText block and opening a new one I avoid the rest of the text being displayed as code-block.

README.md
PhilippMDoerner: Ohhhhh yeah I broke those links, one sec, that should be an easy fix

book/internals/hooks/after_build_hook.nim
PhilippMDoerner: Instinctively my opinion is that only code that adds/fixes features and contributes to owlkettle doing its actual task should be part of the actual source. The actual source should never have to change for things outside of it , such as tests or documentation, as the necessity for such a feature is non-obvious.

Is the issue that you would want to get more guarantees that certain things work by having brew run, execute most of its code and return *just* before it would launch a GUI?

can-lehmann: This is not about guarantees of correctness of the examples, but rather about the readability of the docs. I think you should be able to just copy/paste the examples from the docs without any modification. As a reader of the docs I would ask myself "Why do I have to uncomment this before I can try it out?"

From a software engineering perspective, I think we are long past the point of cleanly separating docs & actual implementation; I mean there is basically a whole documentation generator hidden behind the `-d:owlkettleDocs` flag, so adding another flag would not really hurt. :smile:

PhilippMDoerner:
> I made some minor corrections of the `internals.md` text after merging, those should be applied here as well: [c211d4c](https://github.com/can-lehmann/owlkettle/commit/c211d4c0f9f2d6f8bea5da322ae45ea0ad2e9294)

Thanks for the hint! I really should sync my branches again. I updated the various 15 or so files I split internals.md up into, I also added the changes to widgets.nim and re-compiled the widgets.md from there.

README.md
PhilippMDoerner: Version with fixed links is now deployed for viewing, I'll mark this as resolved

README.md
PhilippMDoerner: Version with fixed links is now deployed for viewing, I'll mark this as resolved

book/internals/hooks/after_build_hook.nim
PhilippMDoerner: If owlkettleDocs is already a thing, mind if we use that very same switch? I see no real benefit in having them separate when they're both for docs in their own way. I'd maybe apply a comment explaining as to why a "when defined(owlkettleDocs): return` is necessary in brew as well

book/internals/adders/adders_with_properties.nim
PhilippMDoerner: The issue appears to have been a trailing whitespace. Not sure if that was a nimibook bug or not, the authors have been informed, and I fixed the thing and removed the workaround. :partying_face:

book/internals/hooks/after_build_hook.nim
PhilippMDoerner: I just skimmed through the brew proc implementations and noticed that we have 3 `brew` procs and their implementations are all different from one another. 

The first generates an `AppConfig` instance and then just runs `runMainLoop`.
The second generates an `AppConfig` instance, an `activateCallback` and then instantiates an application and calls seemingly completely different functions, even though superficially it is identical to the other one with only the "id"-parameter differing
The third is seemingly like the first but with provides a style-manager.

Thus I see no super easy way to implement this cleanly in a single place.
I think putting a `when defined/owlkettleDocs: return` switch could be added to `runMainLoop` ? But that still leaves out the second `brew` implementation which doesn't use `runMainLoop`.  What's the way to go here?

PhilippMDoerner: @can-lehmann I've decided to just go with applying the following constructs to all owlkettle examples:

```nim
when not defined(owlkettleDocs):
  brew(gui(App()))
``` 

That should do

ire4ever1190:
Is there a way to turn off the docs printing in the hooks section?
image

Seems a bit unneeded since you can just read the source given

PhilippMDoerner:
That issue has been fixed, the docs now no longer print as before and can be just copy-pasted and run.
Afaik the nimibook team also has decided to introduce a code block that allows a specific line of code to compile, but not to run.

This can strictly speaking already be done by defining this template:
```nim
template nbCodeOwnkettle(body: untyped) =
  nbCode:
    body
  nb.blk.code &= "brew(...)"
  nb.blk.context["code"] = nb.blk.code
```

But I'd leave that for a future PR. 
@can-lehmann is there anything missing here for you? (I also reverted the links in the README.md to as they were before, I think changing those makes sense to do in a second PR once nimibook pages are deployed)

The only work remaining would need to be done after some kind of go-ahead on your end:
1) After go--ahead
- Add token to repository secrets (that only you can do)
- Remove my branch from the docs.yml 

2) After PR (in a second PR)
- Update README.md to use new links

book/internals/adders.nim
```diff
@@ -0,0 +1,36 @@
+import nimib, nimibook
+
+nbInit(theme = useNimibook)
+nbText: """
+## **Adders**
+
+If you remember the previous example with CustomLabel, not that you weren't able to write:
```
can-lehmann, line 7: This sentence currently does not make sense to me. Maybe change it to something like `In the previous example, you were not able to write the following:`

can-lehmann (commented):
I can't seem to comment on the images in the PR, so I will add the comment here: They are currently rendered in a dark theme, so they look quite weird when using a light theme. One solution might be rendering them in light mode and making the background transparent. This might look ok with both dark & light themes.

book/internals/adders.nim
PhilippMDoerner: Ah, that should've been a "note that you weren't able to write", I'll reword it.

PhilippMDoerner:
> I can't seem to comment on the images in the PR, so I will add the comment here: They are currently rendered in a dark theme, so they look quite weird when using a light theme. One solution might be rendering them in light mode and making the background transparent. This might look ok with both dark & light themes.

I am actually not sure how to technically achieve what you've described here. Like, what I can think of would be doing the screenshot, throwing it into inkscape and then manually cutting all the elements out - or redoing it all in drawio with just a transparent background.

can-lehmann:
Screenshot -> Gimp -> Add alpha to layer -> Autoselect Background -> delete -> re export

can-lehmann:
A general note on this PR: There is not much left to fix on your side as far as I can see. Merging for me is however not that simple, as there are some other factors here:

- **Legal Aspects**: I will have to ensure that all legal requirements are met, publishing (independent of where you host it!) websites from where I live is actually quite involved. This includes looking into whether this requires an Imprint and a GDPR notice (it might not, I am just not sure atm).
- **Security:** I will need to review the security of the workflow at some point. I saw that there is actually an official (as far as I can tell) GitHub action for deploying Pages: https://github.com/actions/deploy-pages It does not seem to use any special access tokens, but rather the built-in [GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication), which provides much more fine grained access control than both the fine grained access tokens and the classic tokens. I will have to do a bit more reading though.

PhilippMDoerner:
<h3> GDPR </h3>

Should not be required:
> If you are collecting data directly from someone, you have to provide them with your privacy notice at the moment you do so.

Source: https://gdpr.eu/privacy-notice/

> Who does the GDPR apply to?
>  All EU organisations that collect, store or otherwise process the personal data of individuals residing in the EU, even if they are not EU citizens.

Source: https://www.itgovernance.eu/en-ie/eu-general-data-protection-regulation-gdpr-ie

Given that I don't remember adding any JS that collects personal information, that should work. 

<h3> Imprint/Impressumspflicht</h3>

I am aware of Imprint requirements in germany given that I stem from there. Based on your awareness of it, I'll go ahead and assume you can at least read german. If you're a commercial organization, which you're not unless you plan on selling owlkettle support for a fee or sth., then it is not required. 

Source 1 in english that is mostly paraphrasing: https://www.iubenda.com/en/help/7816-impressum-what-is-it-and-when-is-it-needed

Source 2 in german directly from the corresponding ministerium: https://www.bmuv.de/themen/verbraucherschutz-im-bmuv/digitaler-verbraucherschutz/impressumspflicht#c66866

> Grundsätzlich kann man sagen, dass die Pflicht für alle Anbieter einer Internetseite gilt, wenn die Plattform geschäftlichen Zwecken dient. Damit fallen ausschließlich privat genutzte Seiten, die sich nur an Familie und Freunde richten, nicht unter die Impressumspflicht. Auch Vereine und Bürgerinitiativen brauchen beispielsweise ein Impressum.

>  Wichtig: Ist auf der privat genutzten Seite Werbung geschaltet, mit der Geld verdient wird, kann dies eine Impressumspflicht begründen.

> Vor allem Verkaufsplattformen wie Online-Shops und Suchmaschinen müssen ein Impressum angeben. Aber auch Accounts in sozialen Netzwerken wie Facebook, Instagram, Twitter und Co. benötigen ein Impressum, wenn das Konto auch gewerblich, beispielsweise für das Bewerben von Produkten oder Stellenanzeigen, genutzt wird. Sie benötigen jedoch kein Impressum, wenn Sie nur gelegentlich privat über Verkaufsplattformen anbieten, also zum Beispiel vereinzelt gebrauchte Schuhe oder das alte Fahrrad im Internet verkaufen.

> Paragraf 5 TMG spricht außerdem von "in der Regel gegen Entgelt angebotene Telemedien". Es genügt also, wenn mit der angebotenen Leistung auf dem Markt Geld verdient werden kann. Der Anbieter muss dies nicht zwingend tun.

Note how this is very much focussing on commercial use where money exchanges hands and you actually have legal entities interacting with one another (a club e.g. needs to interact with its members for fees and to enable collective action and representation, none of that applies here).

So on that front I think we're fine

PhilippMDoerner:
> Screenshot -> Gimp -> Add alpha to layer -> Autoselect Background -> delete -> re export

As for this approach, I have tried an proven incompetent. My image manipulation background is **incredibly** basic. There's a reason I just use drawio. A bit of cropping in inkscape is about the extent of my knowledge.

So I tried the approach, I actually failed at comprehending where and how to add an alpha layer to and it likely would take me a while to just understand the concepts of what that is and how to manipulate it.

I tried converting the mmd files to pngs using the mermaid CLI, but that requires a puppeteer setup that I really don't feel like doing.

I tried just redoing them in drawio, but they just don't hold up in dark mode:

class_diagram
workflow

image
image

I'm... somewhat at an impasse

can-lehmann:
Hmm, you could try adding a bit of custom CSS like this (untested):

```css
.coal > img, .navy > img, .ayu > img {
  filter: invert(100%);
}
```

can-lehmann:
Re Imprint: I did a bit of research and found out that there are basically 2 different laws that govern whether you need an imprint. Either §5 TMG or §18 MStV. You seem to mainly focus on [§5 TMG](https://www.gesetze-im-internet.de/tmg/__5.html), which talks about commerical services. Obiously this does not apply to owlkettle. However [§18 MStV](https://www.die-medienanstalten.de/fileadmin/user_upload/Rechtsgrundlagen/Gesetze_Staatsvertraege/Medienstaatsvertrag_MStV.pdf) might (?) apply to our case. See section 2a here: https://www.lfk.de/fileadmin/PDFs/Dokumente_und_Rechtsgrundlagen/Leitfaeden/leitfaden-impressumspflicht-2022.pdf

can-lehmann:
It is a bit questionable if

> Verfügt das Angebot dagegen über eine große Anzahl an Nutzern, soll es eine breite Öffentlichkeit erreichen oder ist es auf die Generierung möglichst vieler Follower bzw. Abonnenten ausgerichtet [...]
> ~ Leitfaden zur Impressumspflicht im Internet (Section 2(a))

applies here though.

PhilippMDoerner:
https://www.ionos.de/digitalguide/websites/online-recht/impressum-im-internet-was-gilt-es-zu-beachten/
Ionos treats it a bit stricter.

However, I could argue that there already is an impressum available: Githubs own.

PhilippMDoerner:
Further resources: https://law.stackexchange.com/questions/28459/do-github-projects-by-germans-require-an-imprint-link-in-their-readme

This is an example imprint of somebody: https://github.com/johsteffens/Impressum

PhilippMDoerner:
> Hmm, you could try adding a bit of custom CSS like this (untested):
> 
> ```css
> .coal img, .navy img, .ayu img {
>   filter: invert(100%);
> }
> ```
> 
> Edit: Sorry, it needs to be without the `>`. And you would also need some way to make sure that the other images are not affected (e.g. a css class).

That worked well. I did do some additional changes though:
1) I added this trick to the CONTRIBUTING.md so others in the future won't have to research it
2) Unrelated to any of that, I moved the "Contributing"-section in the nimibook back to the bottom.
I originally had the "References" there, because of a nimibook bug with indentation, but that has been fixed recently, so the workaround is no longer necessary.

can-lehmann:
Ah, sorry for the delay. I currently have some irl deadlines coming up and therefore less time to work on private/open source projects. My schedule should have cleared up next week (around thursday). My current plan for moving forward then is the following:

1. Create a 2.2.0 release of the current state of owlkettle, in order to have a working version of the library before we make incompatible changes for 3.0.0
2. Merge this PR
3. Add an imprint & other legal requirements
4. Enable GitHub Pages for the repository

Great work on moving the workflow to the GITHUB_TOKEN btw, at least from the commit history it looks like that was quite involved.

can-lehmann:
So specifically regarding the imprint, to me it seems like we probably need one. I will add that after merging.

PhilippMDoerner:
@can-lehmann 
Since I managed to wrap my head around setters with the most recent PR about CSS classes, I felt so free as to add some setter docs.
They describe setters as a convenience-feature for the most part.
[Here](https://philippmdoerner.github.io/owlkettle/book/internals/setters.html) a link to the docs specifically.

book/internals/setters.nim
```diff
@@ -0,0 +1,76 @@
+import nimib, nimibook
+
+nbInit(theme = useNimibook)
+
+nbText: """
+## **Setters**
+Setters are a convenience mechanism for widgets.
+Through them you can hide fields on your `WidgetState`.
```
can-lehmann, line 8: This might be misleading. The fields can still be accessed.

book/internals/setters.nim
```diff
@@ -0,0 +1,76 @@
+import nimib, nimibook
+
+nbInit(theme = useNimibook)
+
+nbText: """
+## **Setters**
+Setters are a convenience mechanism for widgets.
```
can-lehmann, line 7:
```suggestion
Setters are a convenience mechanism for creating widgets.
```

book/internals/setters.nim
```diff
@@ -0,0 +1,76 @@
...
+Through them you can hide fields on your `WidgetState`.
+Parent-widgets will be able to assign to "pseudo-fields", whose assigned values get converted and then forwarded to the actual field on `WidgetState`.
```
can-lehmann, line 9:
```suggestion
When creating widgets, setters act just like any other fields on the widget. When you assign a value to a setter, it is executed and can decide how the assigned value should be handled.

One example is the `Button.icon` setter. When assigning to it, the setter adds an `Icon` widget with the given name to the `Button`.

Another usecase is converting the type assigned to a field before storing it. For example, the `BaseWidget.margin` setter has overloads for both the `Margin` object and `int`. When you assign an `int` to the `margin` setter, it first converts it to a `Margin` object with the same margin on all sides. When you assign a `Margin` object, it gets stored directly.
```

book/internals/setters.nim
```diff
@@ -0,0 +1,76 @@
...
+
+allow parent-widgets to assign to fields on `WidgetState` with different types of values that get converted to the field on `WidgetState`!
+
+Every setter consists of:
+- a `setter <PseudoFieldName>` declaration with the type that the setter receives, e.g. `setter margins: int` defines a setter that will receive an int from a parent-widget
```
can-lehmann, line 14: I would just write SetterName instead of PseudoFieldName. (applies to all instances of PseudoFieldName).

book/internals/setters.nim
```diff
@@ -0,0 +1,76 @@
...
+      Window:
+        Box(orient = OrientX):
+          MyViewable(margins = [0, 60, 60])
+          MyViewable(margins = 40)
+
+  when not defined(owlkettleNimiDocs):
+    brew(gui(App()))
+
+nbText: """
+This defines a `MyViewable` that has a pseudo-field called `margins`, which hides the "true" field `internalMargins`.
```
can-lehmann, line 65:
```suggestion
This defines a `MyViewable` that has a setter called `margins`, which hides the internal field `internalMargins`.
```

PhilippMDoerner: I agree with your statement that "hiding" is misleading in general. Setters in that sense don't forbid direct assignment to the field, though imo it is strongly implied that using a setter means you should use the setter instead of direct assignment to the field.

Anyway, I wouldn't start talking about "hiding" in general there, so I just rephrased the entire section

book/internals/setters.nim
PhilippMDoerner: Agreed, I removed that mention and rewrote the later section that also talks about "hiding" stuff.
Speaking of "Hiding" in that sense really only makes sense if we either want to explicitly discourage assigning directly to fields with setters, or implement a mechanism that directly blocks it.

can-lehmann:
Looks good. 

Regarding blocking assignment on a field: I don't think this currently possible, though I am planning to add something in this direction for 3.0.0. You can however discourage assigning to a field by marking it as `{.internal.}`, which will hide it from the docs.

PhilippMDoerner:
> Looks good.
> 
> Regarding blocking assignment on a field: I don't think this currently possible, though I am planning to add something in this direction for 3.0.0. You can however discourage assigning to a field by marking it as `{.internal.}`, which will hide it from the docs.

Ohhh should I add that to the setter docs, including a general recommendation of "If you use setters, don't assign to the field that they assign to" ?

can-lehmann (approved):
I think I will begin the process of merging this now.

PhilippMDoerner:
Works for me, I'll make a mini PR for the internals pragma at a later date

can-lehmann:
> Works for me, I'll make a mini PR for the internals pragma at a later date

:+1: I think we should wait for the changes for 3.0.0 here, since there will be some changes regarding the pragmas (including `{.internal.}`) there anyways.

PhilippMDoerner:
Alrighty, works for me!

can-lehmann:
Just merged this, thanks again for all the work you put into it!

PR #38, opened 2023-03-08
Feature/#36 custom css class support

### LINKED ISSUE #36: Add support for setting custom css classes
Wanted to add this issue mostly so we don't forget about it and because it's personally important to me.

Currently, there is no official way to set a custom CSS class to a widget, which you could style as you'd want.

There is an unofficial hacky way that we talked about (documented here for any future reader) - Use the `afterBuild` hook :

```nim
import owlkettle, owlkettle/gtk

viewable RedCounterButton:
  count: int = 0
  
  hooks:
    afterBuild:
      let
        internalWidget = state.unwrapInternalWidget()
        styleContext = gtk_widget_get_style_context(internalWidget)
      gtk_style_context_add_class(styleContext, "red-background")

method view(state: RedCounterButtonState): Widget =
  result = gui:
    Button:
      text = $state.count
      proc clicked() =
        state.count += 1

viewable App:
  discard

method view(app: AppState): Widget =
  result = gui:
    Window:
      RedCounterButton()

brew(gui(App()), stylesheets=["stylesheet.css"])
```
And in `stylesheet.css`:
```css
.red-background {
  background: #ff0000;
}
```

(This only works with the current head commit, current being 25.02.2023, for anyone trying to do this).

I'm honestly not sure how a "proper" solution would be implemented. Should we have some kind of static "cssClass" field on `WidgetState` that is assumed to be there and whose value gets transferred to the internal-widget during the `build` and `update` phases after the hooks have run?

PhilippMDoerner:
NOTE: ONLY MERGE AFTER NIMIBOOK DOCS WERE MERGED

This branch already carries the changes that the nimibook-docs2 branch carries in the other PR. This was necessary, as the nimibook-doc examples provide an additional layer of tests and security that this refactoring didn't just break everything somehow.

#### The actual PR description

This PR introduces the "style" property to BaseWidget and thus to all Widgets that inherit from BaseWidget.
Note, that this does not mean that all **Viewables** have access to it, only Widgets inheriting from BaseWidget.

Question here: If I want to define a custom widget with css-classes, is that even possible? Do they just need to inherit from BaseWidget? Or do they need to add a "Box"-widget first which they can then assign CSS-classes to?

closes #36


owlkettle/widgets.nim
```diff
@@ -36,31 +36,37 @@ type Margin* = object
 renderable BaseWidget:
   ## The base widget of all widgets. Supports redrawing the entire Application
   ## by calling `<WidgetName>State.app.redraw()
+  internalMargin {.internal.}: Margin = Margin() ## Allows setting top, bottom, left and right margin of a widget. Margin has those names as fields to set integer values to.
+  internalStyle: HashSet[StyleClass] = initHashSet[StyleClass]()
```
can-lehmann, line 40:
```suggestion
  internalStyle {.internal.}: HashSet[StyleClass] = initHashSet[StyleClass]()
```

can-lehmann (commented):
Looks good so far, I will do a more detailed review once the nimibook branch is merged. I just remembered that I used the old style field in the tutorial and README. The code snippets there will need to be updated too.

PhilippMDoerner:
As we talked in discord earlier, I'll close this PR as the lack of github recognizing that most of these commits are already merged is a problem.

PR #40, opened 2023-03-15
Css class attempt2

PhilippMDoerner:
Here we go again!

Follow up PR to https://github.com/can-lehmann/owlkettle/pull/38
Hopefully this one works better!


can-lehmann (changes requested):
The examples in the README and tutorial need to be changed to also use `style = [...]` instead of `style = {...}`.

PR #54, opened 2023-04-09
Fix `gorge` breaking builds

RickBarretto:
Basically, I did just two adjusts that permits me compile any `owlkettle` in Windows system.
I don't know if this problem exists in others OSes, but while using `Msys2` I had some problems to compile the example application.

- You can see #53 to more information.

But, in summary, `gorge` returns a multiline string that needs to be stripped.

> Side note: I imported `strutils` using `as` parameter to avoid any variable conflict with the rest of the implementation...

### DIFF
--- a/owlkettle/cairo.nim
+++ b/owlkettle/cairo.nim
@@ -77,7 +77,8 @@ type
 
 proc `==`(a, b: CairoStatus): bool {.borrow.}
 
-{.passl: gorge("pkg-config --libs cairo").}
+import std/strutils as strutils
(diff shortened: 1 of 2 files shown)


can-lehmann:
Thank you!
I was just in the process of writing a comment on your discussion thread asking if you would be willing to create a PR for this. I guess I was too slow :smile:

SolitudeSF:
> > Side note: I imported `strutils` using `as` parameter to avoid any variable conflict with the rest of the implementation...

that doesnt do what you think it does

RickBarretto:
> that doesnt do what you think it does

Doesn't? So what should I do instead?

PR #58, opened 2023-04-19
Update distributing process

RickBarretto:
- add the right way to copy necessary `dll`s from `mingw*/`
- add the right way to copy `share/` and `lib/` from `mingw*/`
- add a minor remind about `libadwaita`

---

Related to:
- #57


can-lehmann:
Looks good. Thanks again for figuring all of this out!

PR #60, opened 2023-04-26
Add Gtk.Window::fullscreened

aleksrutins:
This adds the [fullscreened](https://docs.gtk.org/gtk4/property.Window.fullscreened.html) property to `Window`. It's implemented using `gtk_window_fullscren` and `gtk_window_unfullscreen`.

### DIFF
--- a/owlkettle/widgets.nim
+++ b/owlkettle/widgets.nim
@@ -144,10 +144,8 @@ renderable Window of BaseWindow:
   hooks fullscreened:
     property:
       if state.fullscreened:
-        echo "fullscreening"
         gtk_window_fullscreen(state.internalWidget)
       else:
-        echo "unfullscreening"
         gtk_window_unfullscreen(state.internalWidget)
   
   hooks titlebar:
--- a/owlkettle/widgets.nim
+++ b/owlkettle/widgets.nim
@@ -128,16 +128,13 @@ renderable BaseWindow of BaseWidget:
 
 renderable Window of BaseWindow:
   title: string
-  fullscreened: bool ## Whether or not the window is fullscreened; setting this property is equivalent to calling `gtk_window_fullscreen` and `gtk_window_unfullscreen` for values of `true` and `false`, respectively. Either operation is asynchronous, which means you will need to connect to the `::notify` signal in order to know whether the operation was successful (it could fail if, for instance, a window manager refuses to respond to the signal).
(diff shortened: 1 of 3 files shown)


can-lehmann:
Looks good so far. Let me know when this is ready for review.

aleksrutins:
After trying it in C++ and getting the same exact bug I was trying to debug (fullscreening momentarily and then returning to normal), I think it's a problem with how I'm using it.

owlkettle/widgets.nim
```diff
@@ -128,17 +128,28 @@ renderable BaseWindow of BaseWidget:
 
 renderable Window of BaseWindow:
   title: string
+  fullscreened: bool ## Whether or not the window is fullscreened; setting this property is equivalent to calling `gtk_window_fullscreen` and `gtk_window_unfullscreen` for values of `true` and `false`, respectively. Either operation is asynchronous, which means you will need to connect to the `::notify` signal in order to know whether the operation was successful (it could fail if, for instance, a window manager refuses to respond to the signal).
```
can-lehmann, line 131: The docstring here seems to be copied from the GTK documentation. Since GTK itself it LGPL (and the docs are probably too, not sure), we cannot include it in owlkettle. The provided information also does not apply to owlkettle, since it is impossible to connect to the `::notify` signal in owlkettle. I would just remove the comment entirely.

owlkettle/widgets.nim
```diff
@@ -128,17 +128,28 @@ renderable BaseWindow of BaseWidget:
...
   titlebar: Widget ## Custom widget set as the titlebar of the window
   child: Widget
   
   hooks:
     beforeBuild:
       state.internalWidget = gtk_window_new(GTK_WINDOW_TOPLEVEL)
+    afterBuild:
+      if state.fullscreened:
+        gtk_window_fullscreen(state.internalWidget)
```
can-lehmann, line 140: Why is this required?

aleksrutins: That was some confused debugging that I forgot to remove; I'll take it out.

owlkettle/widgets.nim
aleksrutins: Ah, I didn't realize that. I'll remove it.

can-lehmann:
Looks good!

PR #72, opened 2023-09-10
Add spinner widget

SirOlaf:
Nice for showing that something is loading/processing.


can-lehmann:
Looks good! Could you run `make -C docs` to update the `widgets.md` widget reference?

Issue #75, opened 2023-09-13
Update `=destroy` hooks


can-lehmann:
With Nim 2.0.0, the function signature for `=destroy` hooks changed.
These need to be updated, the compiler warnings should tell you the relevant locations.

can-lehmann:
@PhilippMDoerner do you want me to assign you to this?

PhilippMDoerner:
Yeah, might make for good research ~~and when I'm too bored at the prospect of updating docs of my mapster package this might be a good fit~~

PhilippMDoerner:
From my understanding this only affects 2 places of code:
1) PixiBuf in widgets.nim around line ~460
```nim
  PixbufObj* = object
    gdk: GdkPixbuf
  
  PixBuf* = ref PixbufObj

proc `=destroy`(x: PixbufObj) =
  echo "PixbufObj destructor"
  g_object_unref(pointer(x.gdk))

proc newPixbuf(gdk: GdkPixbuf): Pixbuf =
  if gdk.isNil:
    raise newException(ValueError, "Unable to create Pixbuf from GdkPixbuf(nil)")
  new(result)
  result.gdk = gdk
  ```
2) TextBufferObj in widgets.nim around line ~2000

```nim
proc `=destroy`(buffer: TextBufferObj) =
  echo "DESTRUCTOR TEXT"
  g_object_unref(pointer(buffer.gtk))

proc newTextBuffer*(): TextBuffer =
  new(result)
  result.gtk = gtk_text_buffer_new(nil.GtkTextTagTable)
```

You can test 1) by compiling and playing around with examples/widgets/picture.nim
You'll be able to see how it triggers the echo.

No clue about TextBufferObj as of now and what else would need to happen to avoid memory leaks in the case of exceptions occurring during destruction.

can-lehmann:
Looks good so far, I am just going to trust that you tested it :)
The text buffer should be testable using the text view example.

If it fixes the warnings, it should also close this issue. I don't see anything about **recoverable** errors in the g_object_unref docs, so that should be fine. The minimum Nim version requirement in the nimble file (and the workflows) needs to be raised to 2.0.0, because I am not sure if previous Nim versions will execute these destructors. If you want you could also add the destructor for `Stylesheet` in the same PR (https://github.com/can-lehmann/owlkettle/issues/76).

PhilippMDoerner:
I tested only PixBufObj so far given I haven't found an example to compile where textbuffer is used.
[...]

PR #81, opened 2023-09-15
#75 destroy hooks

PhilippMDoerner:
This is a branch containing 2 approaches to the entire destructor semantics:
1) (First commit) Use a value object to wrap GdkPixBuf called PixBuf
2) (Second commit) use a type alias to GdkPixBuf called PixBuf directly without any value objects.

To go back to the first attempt, just revert the second commit.

### DIFF
--- a/owlkettle.nimble
+++ b/owlkettle.nimble
@@ -18,7 +18,7 @@ task examples, "Build examples":
   withDir "examples":
     for file in findExamples("."):
       echo "INFO: Compile " & file
-      exec "nim c --hints:off --verbosity:0 " & file
+      exec "nim c --hints:off --path:.. --verbosity:0 " & file
       echo "INFO: OK"
       echo "================================================================================"
 
--- a/owlkettle/customPragmas.nim
+++ b/owlkettle/customPragmas.nim
@@ -1,0 +1,8 @@
+## Definess a custom pragma called "locker".
+## This does nothing when compiled with a nim version 2.0 or higher,
+## but applies the "locks: 0" pragma if compiled for e.g. 1.6.X.
+## "locks: 0" is applied to ensure that this code never accesses locked data in user-defined procs.
+when NimMajor >= 2:
+  {.pragma: locker.}
+else:
+  {.pragma: locker, locks: 0.}
--- a/owlkettle/customPragmas.nim
+++ b/owlkettle/customPragmas.nim
@@ -1,4 +1,0 @@
-## Definess a custom pragma called "locker".
-## This does nothing when compiled with a nim version 2.0 or higher,
-## but applies the "locks: 0" pragma if compiled for e.g. 1.6.X.
-## "locks: 0" is applied to ensure that this code never accesses locked data in user-defined procs.
--- a/owlkettle/customPragmas.nim
+++ b/owlkettle/customPragmas.nim
@@ -2,7 +2,3 @@
 ## This does nothing when compiled with a nim version 2.0 or higher,
 ## but applies the "locks: 0" pragma if compiled for e.g. 1.6.X.
 ## "locks: 0" is applied to ensure that this code never accesses locked data in user-defined procs.
-when NimMajor >= 2:
-  {.pragma: locker.}
-else:
-  {.pragma: locker, locks: 0.}
--- a/owlkettle/gtk.nim
+++ b/owlkettle/gtk.nim
@@ -340,6 +340,7 @@ proc g_signal_connect_data*(widget: pointer,
 proc g_type_from_name*(name: cstring): GType
 proc g_object_new*(typ: GType): pointer {.varargs.}
 proc g_object_ref*(obj: pointer)
+proc g_object_weak_ref*(obj: pointer, notify: proc (data, oldObjectPtr: pointer) {.cdecl.}, data: pointer)
 proc g_object_unref*(obj: pointer)
 proc g_object_set_property*(obj: pointer, name: cstring, value: ptr GValue)
 proc g_type_fundamental*(id: GType): GType
--- a/owlkettle/mainloop.nim
+++ b/owlkettle/mainloop.nim
@@ -20,7 +20,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-import gtk, widgetdef
+import gtk, widgetdef, common
 
 type 
   StylesheetObj = object
@@ -29,18 +29,12 @@ type
 
   Stylesheet* = ref StylesheetObj
 
-when(NimMajor >= 2):
-  proc `=destroy`(stylesheet: StylesheetObj) =
+crossVersionDestructor(stylesheet, StylesheetObj):
     if isNil(stylesheet.provider):
       return
     
     g_object_unref(pointer(stylesheet.provider))
-else:
-  proc `=destroy`(stylesheet: var StylesheetObj) =
-    if isNil(stylesheet.provider):
-      return
-    
-    g_object_unref(pointer(stylesheet.provider))
+
 proc `=copy`*(dest: var StylesheetObj, source: StylesheetObj) =
   let areSameObject = pointer(source.provider) == pointer(dest.provider)
   if areSameObject:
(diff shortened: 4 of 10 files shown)


can-lehmann:
Adding a `=sink` hook fixes it for me. I have no idea why though. I plan to investigate further when I have some more time.

owlkettle/gtk.nim
```diff
@@ -340,6 +340,7 @@ proc g_signal_connect_data*(widget: pointer,
 proc g_type_from_name*(name: cstring): GType
 proc g_object_new*(typ: GType): pointer {.varargs.}
 proc g_object_ref*(obj: pointer)
+proc g_object_weak_ref*(obj: pointer, notify: proc (data, oldObjectPtr: pointer) {.cdecl.}, data: pointer)
```
PhilippMDoerner, line 343: I feel tempted to leave this there just to have it as it shouldn't hurt, disagreement here?

PhilippMDoerner:
We're running into 
```txt
Error: 403 - {"message":"API rate limit exceeded for 13.88.98.81. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```
The tests themselves *should* run, at least they work for me on 2.0 locally.
However I also modified the path used.

owlkettle/gtk.nim
can-lehmann: Yes, I would also leave it. It might be useful for debugging in the future.

owlkettle/customPragmas.nim
```diff
@@ -0,0 +1,8 @@
+## Definess a custom pragma called "locker".
+## This does nothing when compiled with a nim version 2.0 or higher,
+## but applies the "locks: 0" pragma if compiled for e.g. 1.6.X.
+## "locks: 0" is applied to ensure that this code never accesses locked data in user-defined procs.
+when NimMajor >= 2:
+  {.pragma: locker.}
+else:
+  {.pragma: locker, locks: 0.}
```
can-lehmann, line 8: I think this would nicely fit into `common.nim`

can-lehmann:
I think we can move forward with 3. for now. You could use a `template` to reduce code duplication (put that into `common.nim` as well).

owlkettle/customPragmas.nim
PhilippMDoerner: Based on your lower comment I assume you want to move this into a `customPragma` template and use that in the various modules?

can-lehmann: Oh, sorry, no, I meant putting the `=destroy` in a template for supporting both 1.6.x and 2.0.0. But if a template is need here, using one is also fine.

PhilippMDoerner:
> I think we can move forward with 3. for now. You could use a `template` to reduce code duplication (put that into `common.nim` as well).

I'm actually somewhat torn here.
Because as far as I can see judge freeing the memory for a Pixbuf is a different task than freeing the memory of a TextBuffer is a different task than freeing the memory of a Cstring.
At least for now that's what it looks like to me.
Given that they don't seem inherently coupled to me and it just so happens that their destructors look similar, I hesitate to start coupling some of them together (because cstrings will most likely need their own strategy) when they should be different things.

Feel free to overrule me here, that's just my current instinct from my very limited experience with owlkettle and GTK.

can-lehmann:
I agree, I would do something like:

```nim
template destructor(name, typ, body: untyped) =
  when(NimMajor >= 2):
    proc `=destroy`(name: typ) =
      body
  else:
    proc `=destroy`(name: var typ) =
      body

destructor(pixbuf, Pixbuf):
  ...
```

owlkettle/customPragmas.nim
can-lehmann, line 1: This file needs to be deleted from git.

owlkettle/widgets.nim
```diff
@@ -1995,11 +2032,15 @@ type
...
+when(NimMajor >= 2):
+  proc `=destroy`(buffer: TextBufferObj) =
+    g_object_unref(pointer(buffer.gtk))
+else:
+  proc `=destroy`(buffer: var TextBufferObj) =
+    g_object_unref(pointer(buffer.gtk))
 
 proc newTextBuffer*(): TextBuffer =
-  new(result, finalizer=finalizer)
+  new(result)
```
can-lehmann, line 2043:
```suggestion
  result = TextBuffer(gtk: gtk_text_buffer_new(nil.GtkTextTagTable))
```

owlkettle/widgets.nim
```diff
@@ -1995,11 +2032,15 @@ type
...
   result.gtk = gtk_text_buffer_new(nil.GtkTextTagTable)
```
can-lehmann, line 2044:
```suggestion

```

owlkettle/widgets.nim
```diff
@@ -458,18 +459,48 @@ type
...
+  if not isNil(source.gdk):
+    g_object_ref(pointer(source.gdk))
+    
+  dest.gdk = source.gdk
+    
 proc newPixbuf(gdk: GdkPixbuf): Pixbuf =
   if gdk.isNil:
     raise newException(ValueError, "Unable to create Pixbuf from GdkPixbuf(nil)")
-  new(result, finalizer=finalizer)
+  result = Pixbuf()
```
can-lehmann, line 501:
```suggestion
  result = Pixbuf(gdk: gdk)
```

owlkettle/widgets.nim
```diff
@@ -458,18 +459,48 @@ type
...
   result.gdk = gdk
```
can-lehmann, line 502:
```suggestion

```

PhilippMDoerner:
Applied the template to TextBuf, PixBuf and Stylesheet and streamlined the two creation procs for textbuf and pixbuf as suggested.

can-lehmann:
Are the `=copy` hooks still required? They are currently applied inconsistently.

owlkettle/mainloop.nim
```diff
@@ -20,11 +20,32 @@
...
+    provider: GtkCssProvider
+    priority: int
+
+  Stylesheet* = ref StylesheetObj
+
+crossVersionDestructor(stylesheet, StylesheetObj):
+    if isNil(stylesheet.provider):
+      return
+    
+    g_object_unref(pointer(stylesheet.provider))
```
can-lehmann, line 36:
```suggestion
  if isNil(stylesheet.provider):
    return
  
  g_object_unref(pointer(stylesheet.provider))
```

PhilippMDoerner:
> Are the `=copy` hooks still required? They are currently applied inconsistently.

Logically speaking they *should* be required, because a copy means the ref count for gtk goes up by one and they must be informed of that so that they can handle memory management.
I swear I wrote a copy-hook for textbuf at one point, must've gotten deleted over the course of the PR.

can-lehmann:
Since it is a `ref object` though, there should not be any copys (only if you do `pixbuf1[] = pixbuf2[]`).

PhilippMDoerner:
Which is a fair point but in that case I'd still want to have the copy-hook, just for "correctness" even if nobody should ever do that.
So yeah, give me a sec I'll provide a copy hook

owlkettle.nimble
```diff
@@ -18,7 +18,7 @@ task examples, "Build examples":
   withDir "examples":
     for file in findExamples("."):
       echo "INFO: Compile " & file
-      exec "nim c --hints:off --verbosity:0 " & file
+      exec "nim c --hints:off --path:.. --verbosity:0 " & file
```
can-lehmann, line 21: Why is this required?

PhilippMDoerner: I currently very intentionally have no global version of owlkettle installed to make sure when compiling this it **has** to use the local version. If I don't provide a path to this local version (aka the owlkettle **in this repo** instead of the installed owlkettle version) I get:
```nim
/home/philipp/dev/owlkettle/examples/introduction.nim(23, 8) Error: cannot open file: owlkettle
```

Which indicates to me that if you run `nimble examples` as "testing" locally, it'll not use your local owlkettle version.
--path:.. allows you to specify that it shall add wherever your owlkettle repo is to the path, where it can find and thus import owlkettle.nim

This **should** guarantee that it uses the owlkettle files of your repository rather than of whatever owlkettle version you got installed.

can-lehmann: Ah, makes sense.

can-lehmann:
Looks good!

Issue #76, opened 2023-09-13
Missing destructors and memory deallocations


can-lehmann:
Review code for missing destructors and potential memory leaks. Known cases so far are

- GTK functions that return `cstring` (?)

PhilippMDoerner:
Regarding the brew procedure with id.
It has a `  defer: g_object_unref(app.pointer)` line which unrefs the pointer of the application, no (owlkettle.nim line 214)?
So doesn't that mean that the brew procedure does unref things?

PhilippMDoerner:
As for destructors for cstring returning procs... how would that one work in particular?
Create a type `Type ManagedCstring = distinct cstring` and define destructors for that and see about that being used everywhere instead of cstring?

can-lehmann:
> Regarding the brew procedure with id. It has a ` defer: g_object_unref(app.pointer)` line which unrefs the pointer of the application, no (owlkettle.nim line 214)? So doesn't that mean that the brew procedure does unref things?

Oh you are right, I did not see that.

can-lehmann:
> As for destructors for cstring returning procs... how would that one work in particular? Create a type `Type ManagedCstring = distinct cstring` and define destructors for that and see about that being used everywhere instead of cstring?

Well if value semantics worked, that would be the way to go. There are two options:

1. Hope that `=destroy` gets called correctly and use `distinct cstring`.
2. Manually deallocate.

You also **have** to check for each function's docs for whether the ownership of data returned by it is actually transferred to the caller. E.g. [here](https://docs.gtk.org/gtk4/method.Editable.get_text.html) this is not the case and you **must not** call free.

PhilippMDoerner:
In that case that seems like a larger task and I'd like to split out dealing with cstring destructors in a separate PR

PhilippMDoerner:
Okay, I've identified the cstring procs:
```txt
  gtk.nim 350: proc g_value_get_string*(value: ptr GValue): cstring
  gtk.nim 400: proc g_file_get_path*(file: GFile): cstring
  gtk.nim 401: proc g_file_get_uri*(file: GFile): cstring
  gtk.nim 536: proc gtk_widget_get_name*(widget: GtkWidget): cstring
  gtk.nim 598: proc gtk_editable_get_text*(entry: GtkWidget): cstring
    Do not manage memory, you do not own it (See: https://docs.gtk.org/gtk4/method.Editable.get_text.html)
[...]

PR #84, opened 2023-09-17
#76 add missing destructors

PhilippMDoerner:
Adds destructors to the cstring-returning procs from gtk defined in the table in one of my comments in #76

This is a first attempt

### DIFF
--- a/owlkettle/gtk.nim
+++ b/owlkettle/gtk.nim
@@ -365,7 +365,7 @@ proc g_value_unset*(value: ptr GValue)
 proc g_idle_add_full*(priority: cint, fn: GSourceFunc, data: pointer, notify: GDestroyNotify): cuint
 proc g_timeout_add_full*(priority: cint, interval: cuint, fn: GSourceFunc, data: pointer, notify: GDestroyNotify): cuint
 proc g_strdup*(str: cstring): pointer
-
+proc g_free*(str: pointer)
 # GLib.Source
 proc g_source_remove*(id: cuint): cbool
 
--- a/owlkettle/gtk.nim
+++ b/owlkettle/gtk.nim
@@ -370,6 +370,7 @@ proc g_free*(str: pointer)
 {.pop.}
 
 type OwnedGtkString* = distinct cstring
+
   # Forward declarations are needed as otherwise the line `cStr.cstring.isNil` 
   # in the =destroy hook causes nim to use cstring's hooks implicitly, which 
   # we do not want to ensure memory safe use of OwnedGtkString
--- a/owlkettle/gtk.nim
+++ b/owlkettle/gtk.nim
@@ -23,6 +23,7 @@
 # Bindings for GTK 4
 
 import std/[os]
+import ./common
 
 import std/strutils as strutils
 {.passl: strutils.strip(gorge("pkg-config --libs gtk4")).}
@@ -366,6 +367,20 @@ proc g_idle_add_full*(priority: cint, fn: GSourceFunc, data: pointer, notify: GD
 proc g_timeout_add_full*(priority: cint, interval: cuint, fn: GSourceFunc, data: pointer, notify: GDestroyNotify): cuint
 proc g_strdup*(str: cstring): pointer
 proc g_free*(str: pointer)
+{.pop.}
+
+type OwnedGtkString* = distinct cstring
+
+crossVersionDestructor(cstr, OwnedGtkString):
+  echo "Destroy owned gtk string"
+  if cStr.cstring.isNil():
+    return
+  
+  g_free(pointer(cStr))
+
+proc `$`*(cStr: OwnedGtkString): string = $(cStr.cstring)
+
+{.push importc, cdecl.}
 # GLib.Source
 proc g_source_remove*(id: cuint): cbool
 
@@ -401,8 +416,8 @@ proc g_application_run*(app: GApplication, argc: cint, argv: cstringArray): cint
 # Gio.File
 proc g_file_new_for_path*(path: cstring): GFile
 proc g_file_read*(file: GFile, cancelable: pointer, error: ptr GError): GInputStream
-proc g_file_get_path*(file: GFile): cstring
-proc g_file_get_uri*(file: GFile): cstring
+proc g_file_get_path*(file: GFile): OwnedGtkString
+proc g_file_get_uri*(file: GFile): OwnedGtkString
 
 # Gio.InputStream
 proc g_input_stream_close*(stream: GInputStream, cancelable: pointer, error: ptr GError): cbool
@@ -783,7 +798,7 @@ proc gtk_text_buffer_delete*(buffer: GtkTextBuffer, a, b: ptr GtkTextIter)
 proc gtk_text_buffer_set_text*(buffer: GtkTextBuffer, text: cstring, len: cint)
 proc gtk_text_buffer_get_text*(buffer: GtkTextBuffer,
                                a, b: ptr GtkTextIter,
-                               includeHiddenChars: cbool): cstring
+                               includeHiddenChars: cbool): OwnedGtkString
 proc gtk_text_buffer_begin_user_action*(buffer: GtkTextBuffer)
 proc gtk_text_buffer_end_user_action*(buffer: GtkTextBuffer)
 proc gtk_text_buffer_get_start_iter*(buffer: GtkTextBuffer, iter: ptr GtkTextIter)
--- a/owlkettle/gtk.nim
+++ b/owlkettle/gtk.nim
@@ -398,7 +398,7 @@ proc `=sink`*(dest: var OwnedGtkString; source: OwnedGtkString) =
   echo "Sink owned string"
   `=destroy`(dest)
   wasMoved(dest)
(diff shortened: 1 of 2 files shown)


owlkettle/gtk.nim
```diff
@@ -365,7 +366,21 @@ proc g_value_unset*(value: ptr GValue)
 proc g_idle_add_full*(priority: cint, fn: GSourceFunc, data: pointer, notify: GDestroyNotify): cuint
 proc g_timeout_add_full*(priority: cint, interval: cuint, fn: GSourceFunc, data: pointer, notify: GDestroyNotify): cuint
 proc g_strdup*(str: cstring): pointer
+proc g_free*(str: pointer)
+{.pop.}
+
+type OwnedGtkString* = distinct cstring
+
+crossVersionDestructor(cstr, OwnedGtkString):
```
PhilippMDoerner, line 374: It appears I do not necessarily need a =copy hook, because cstring apparently has one and by using cStr.cstring.isNil it implicitly associated the copy-hook from cstring with OwnedGtkString.

owlkettle/gtk.nim
```diff
@@ -401,8 +416,8 @@ proc g_application_run*(app: GApplication, argc: cint, argv: cstringArray): cint
 # Gio.File
 proc g_file_new_for_path*(path: cstring): GFile
 proc g_file_read*(file: GFile, cancelable: pointer, error: ptr GError): GInputStream
-proc g_file_get_path*(file: GFile): cstring
-proc g_file_get_uri*(file: GFile): cstring
+proc g_file_get_path*(file: GFile): OwnedGtkString
```
PhilippMDoerner, line 419: I managed to test g_file_get_path, but not g_file_get_uri because it gets used nowhere.
`gtk_text_buffer_get_text` I couldn't test either because while it gets used in examples/widgets/text_view.nim, it appears to never actually free the string in the textbuffer, ever. Or this proc is never called in there, I don't quite know

can-lehmann: Sounds good!

owlkettle/gtk.nim
can-lehmann: Are you sure that the `=copy` hook is even created at all? (what happens if you `var a = <some owned gtk string>; var b = a; echo $a; echo $b`) If it uses the `cstring` copy hook, that sounds like it might be bad.

PhilippMDoerner: I was under the impression the following inim example would save me:
```nim
nim> type A = distinct cstring
nim> let x: A = "lala".A
nim> let y = x
nim> echo x.pointer.repr
000055870C2332B0
nim> echo y.pointer.repr
000055AA38BA02B0
```

Which is basically the scenario we're running.
Turns out that's false because inim works in different ways than I thought. So we're still boned with the damn copy hook.
The entire approach of forward declaring it might work before the pop situation can screw us, but I'm still running into the scenario that to copy a cstring I have to do so without using `=` and how I'm supposed to do that is beyond me.

PhilippMDoerner: Actually... there is copy behaviour, you get pointer copies, not value copies. Isn't that kinda still okay behaviour?
What is the value of a copy-proc here?

can-lehmann: Pointer copies don't work in this case, as now the `=destroy` hook is called twice on the same pointer. Which is a classic double free error/mistake.

PhilippMDoerner: Okay I think I managed to write a copy-hook that works, at least when I compile this minimal example on my end I get no memory corruption, nor segfault, nor recursive nonsense etc., and the pointers point to different memory locations:
```nim
import std/strutils as strutils
{.passl: strutils.strip(gorge("pkg-config --libs gtk4")).}

{.push importc, cdecl.}
proc g_free*(str: pointer)
proc g_strdup*(str: cstring): pointer
proc g_strnfill*(length: uint64, fill_char: char ): pointer
{.pop.}

type OwnedGtkString* = distinct cstring

proc `=destroy`*(cStr: OwnedGtkString)
proc `=copy`*(dest: var OwnedGtkString, source: OwnedGtkString)

proc `isNil`(cStr: OwnedGtkString): bool {.borrow.}

proc `=destroy`*(cStr: OwnedGtkString) =
  echo "Destroy owned gtk string: ", $cstr.cstring
  if cStr.cstring.isNil():
    return
  
  g_free(pointer(cStr))

proc `=copy`*(dest: var OwnedGtkString, source: OwnedGtkString) =
  let areSameObject = pointer(source) == pointer(dest)
  if areSameObject:
    return
  
  `=destroy`(dest)
  wasMoved(dest)
  let copyC: pointer = g_strdup(cstring(source))
  cstring(dest) = cast[cstring](copyC)


let x: OwnedGtkString = cast[cstring](g_strnfill(5, 'x')).OwnedGtkString
let y = x
echo x.pointer.repr
echo y.pointer.repr
```

can-lehmann: Looks good so far. I would do a `=sink` by following the pattern here: https://nim-lang.org/docs/destructors.html#lifetimeminustracking-hooks-nimeqsink-hook, just to be sure, because of the weird bug we encountered with `distinct` types and hooks.

PhilippMDoerner: I've done so and pushed it, but I can't really test it.
Well, I'm trying to, but I'm struggling with manipulating the code to do what I want here, at least I can't manage to confirm that the copy hook gets called in the picture.nim example

owlkettle/gtk.nim
```diff
@@ -365,7 +366,42 @@ proc g_value_unset*(value: ptr GValue)
 proc g_idle_add_full*(priority: cint, fn: GSourceFunc, data: pointer, notify: GDestroyNotify): cuint
 proc g_timeout_add_full*(priority: cint, interval: cuint, fn: GSourceFunc, data: pointer, notify: GDestroyNotify): cuint
 proc g_strdup*(str: cstring): pointer
+proc g_free*(str: pointer)
+{.pop.}
+
+type OwnedGtkString* = distinct cstring
+  # Forward declarations are needed as otherwise the line `cStr.cstring.isNil`
```
PhilippMDoerner, line 373: Added this comment because imo the reason we do this forward declaration is not straightforward and should be explained.

owlkettle/gtk.nim
```diff
@@ -365,7 +366,42 @@ proc g_value_unset*(value: ptr GValue)
 proc g_idle_add_full*(priority: cint, fn: GSourceFunc, data: pointer, notify: GDestroyNotify): cuint
 proc g_timeout_add_full*(priority: cint, interval: cuint, fn: GSourceFunc, data: pointer, notify: GDestroyNotify): cuint
 proc g_strdup*(str: cstring): pointer
+proc g_free*(str: pointer)
+{.pop.}
+
+type OwnedGtkString* = distinct cstring
+  # Forward declarations are needed as otherwise the line `cStr.cstring.isNil` 
+  # in the =destroy hook causes nim to use cstring's hooks implicitly, which 
+  # we do not want to ensure memory safe use of OwnedGtkString
```
can-lehmann, line 375:
```suggestion

# Forward declarations are needed as otherwise the line `cStr.cstring.isNil` 
# in the =destroy hook causes nim to use cstring's hooks implicitly, which 
# we do not want to ensure memory safe use of OwnedGtkString
```

owlkettle/gtk.nim
```diff
@@ -365,7 +366,42 @@ proc g_value_unset*(value: ptr GValue)
...
+  
+  `=destroy`(dest)
+  wasMoved(dest)
+  cstring(dest) = cast[cstring](g_strdup(cstring(source)))
+
+proc `=sink`*(dest: var OwnedGtkString; source: OwnedGtkString) =
+  echo "Sink owned string"
+  `=destroy`(dest)
+  wasMoved(dest)
+  dest = source
```
can-lehmann, line 400:
```suggestion
  pointer(dest) = pointer(source)
```

can-lehmann: We want to move here, not `=copy`.

PhilippMDoerner: Good point, my lack of being used to deal with hooks and thinking their effects through is showing =/, done

can-lehmann: Oh, sorry, you might need to use `cstring` instead of pointer.

PhilippMDoerner: Huh, that no longer compiles because it really doesn't like assigning to pointer(dest)

PhilippMDoerner: That compiles, whether it actually is functional though I still can't evaluate as I can't even imagine what the code would need to look like that'd show me that it works.

can-lehmann:
I think that we can merge this after removing the `echo` calls. Also the indentation of the comment is still weird.

can-lehmann:
Looks good :+1:

PhilippMDoerner:
Removed the echo's and the comment indentation. That one I originally had in there to not disrupt your organization via comments that used them to divide your bindings into sections based on the file the bound symbol came from.

Note that I still didn't test that the =sink hook gets called anywhere because I couldn't come up with the code for that. Did manage a copy though.

Issue #83, opened 2023-09-16
Add `Scale` widget


can-lehmann:
Gtk's `Scale` widget is currently missing. See https://docs.gtk.org/gtk4/class.Scale.html

PhilippMDoerner:
I've been taking a look at this out of curiosity:
I don't quite get how the hooks know how to connect signals.

For example here:
```nim
renderable Entry of BaseWidget:
  text: string
  placeholder: string ## Shown when the Entry is empty.
  width: int = -1
  maxWidth: int = -1
  xAlign: float = 0.0
  visibility: bool = true
  invisibleChar: Rune = '*'.Rune

  proc changed(text: string) ## Called when the text in the Entry changed
  proc activate() ## Called when the user presses enter/return

  hooks:
    beforeBuild:
      state.internalWidget = gtk_entry_new()
    connectEvents:
      proc changedCallback(widget: GtkWidget, data: ptr EventObj[proc (text: string)]) {.cdecl.} =
        let text = $gtk_editable_get_text(widget)
        EntryState(data[].widget).text = text
        data[].callback(text)
        data[].redraw()
      
      state.connect(state.changed, "changed", changedCallback)
      state.connect(state.activate, "activate", eventCallback)
    disconnectEvents:
      state.internalWidget.disconnect(state.changed)
      state.internalWidget.disconnect(state.activate)
```
Owlkettle knows based on the fact that procs were defined on the renderable that there's a `changed` and an `activate` "field" on the type which can be connected to the GTK signals so that these procs get called.

But what are "changedCallback" and "eventCallback" and how does one know when to use which?

can-lehmann:
The last parameter refers to the proc that is passed directly to GTK for handling the event. Its job is to:

1. Read back any state changes from GTK
2. Call the owlkettle event handler (and pass any arguments it might have)
3. Call `redraw` to update the UI after the event was handled.

E.g. for the changed event, which is called when the user input some text, we need to read back the new text and update the `Entry.text` field with it. Then we call owlkettle's `Entry.changed` handler and finally we call `redraw` to update the UI. `changedCallback` is responsible for doing that.

Other events like `activate` are simpler, because we do not need to read back state & pass arguments to the owlkettle event handler. This is why there is also a general handler called `eventCallback`.

In case of the `Scale` widget, you will need to implement a custom `valueChangedCallback`, that actually reads back the new position of the scale.

PhilippMDoerner:
[...]

PR #90, opened 2023-09-29
#83 add scale widget

PhilippMDoerner:
First draft of the Scale Widget and a single example.
I'd first debate whether it is fine as is and once we have I'd add examples to the widget file


owlkettle/widgets.nim
```diff
@@ -3401,6 +3401,109 @@ renderable AboutDialog of BaseWidget:
...
+  max: float64 = 100
+  inverted: bool = false
+  showValue: bool = true
+  stepSize: float64 = 5
+  pageSize: float64 = 10
+  value: float64 = 0
+  orient: Orient = OrientX
+  showFillLevel: bool = true
+  precision: int64 = 1
+  valuePosition: ScalePosition
```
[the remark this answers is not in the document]
can-lehmann: Also I would personally put `value`, `min` and `max` at the top, because they are probably the most important fields here. → The order the fields appear in the docs is the same as the order they are declared in. So this will improve documentation readability.

owlkettle/widgets.nim
```diff
@@ -3401,6 +3401,109 @@ renderable AboutDialog of BaseWidget:
...
+type ScalePosition* = enum
+  ScaleLeft,
+  ScaleRight,
+  ScaleTop,
+  ScaleBottom
+
+proc toGtk(pos: ScalePosition): GtkPositionType =
+  result = GtkPositionType(ord(pos))
+
+type Mark* = tuple
```
can-lehmann, line 3413: `Mark` is a very generic name (especially since it is exported). Maybe `ScaleMark`?

can-lehmann:
Some general notes:

- Before merging you need to add your example to the `examples/README.md` file. You also have to add a screenshot for it. Ideally this screenshot is taken at 200% screen resolution (see gnome setting / displays), but displayed at half its width, so it looks good on devices with high screen resolutions. (use the existing entries in this file as a guide)
- Regarding code style: See [the style guide](https://github.com/can-lehmann/owlkettle/blob/main/CONTRIBUTING.md#style-guide), mainly the following:
> Imported Procedures: Procedures imported from C code using {.importc.} should always be called using myProc(argument0, ...).

owlkettle/widgets.nim
```diff
@@ -3459,10 +3459,9 @@ renderable Scale of BaseWidget:
       state.value = state.internalWidget.gtk_range_get_value().float64
   
   hooks marks:
-    (build, update):
-      state.internalWidget.gtk_scale_clear_marks()
-      let hasScaleMarks = widget.hasMarks and widget.valMarks.len > 0
-      if hasScaleMarks:
+    property:
+      if widget.hasMarks:
```
can-lehmann, line 3463: property hooks should never access has/valMarks. Use state.marks instead,

owlkettle/widgets.nim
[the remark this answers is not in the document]
PhilippMDoerner: I'm honestly wondering if "pageSize" should be available. Should I just provide a "multiplier" over stepSize (since that's the intention to use it as a larger "step" than the normal one) ?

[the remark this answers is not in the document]
can-lehmann: I think it is generally preferred to stay closer to the GTK widget, so users can just look at the GTK docs.

examples/widgets/scale.nim
```diff
@@ -0,0 +1,71 @@
...
+          precision = 0
+          orient = app.orient
+          marks = marks
+          valuePosition = app.pos
+          
+          proc valueChanged(newValue: float64) =
+            app.value = newValue
+            echo "New value from Scale is ", $newValue
+        
+        Button:
```
can-lehmann, line 57: Add a text / icon & put this in the header bar.

owlkettle/widgets.nim
PhilippMDoerner: Done

owlkettle/widgets.nim
PhilippMDoerner: Done-ish. Basically I just removed the if-check which makes it as correct as it'll get.

examples/widgets/scale.nim
PhilippMDoerner: Done

owlkettle/widgets.nim
```diff
@@ -3401,6 +3401,113 @@ renderable AboutDialog of BaseWidget:
...
+
+  hooks marks:
+    property:
+      state.internalWidget.gtk_scale_clear_marks()
+      for mark in widget.valMarks:
+        let label: string = if mark.label.isSome(): mark.label.get() else: $mark.value
+        gtk_scale_add_mark(state.internalWidget, mark.value , mark.position.toGtk(), label.cstring)
+
+  hooks value:
+    build:
```
PhilippMDoerner, line 3471: I had to add this one back in. If I don't have this build hook, the initial value just doesn't get properly set for value, at least not in my compiles. Can you check if that's also true on your end?

PhilippMDoerner:
Actually do you have a particular feeling about `float` vs `float64`?
Mostly because float implicitly is float64 on basically any reasonably modern system and I don't think I consistently used `float64` everywhere when typing things.

PhilippMDoerner:
> Some general notes:
> 
>     * Before merging you need to add your example to the `examples/README.md` file. You also have to add a screenshot for it. Ideally this screenshot is taken at 200% screen resolution (see gnome setting / displays), but displayed at half its width, so it looks good on devices with high screen resolutions. (use the existing entries in this file as a guide)
> 
>     * Regarding code style: See [the style guide](https://github.com/can-lehmann/owlkettle/blob/main/CONTRIBUTING.md#style-guide), mainly the following:
> 
> 
> > Imported Procedures: Procedures imported from C code using {.importc.} should always be called using myProc(argument0, ...).
> 
>     * Once everything is finished, please rebuild the widget docs (see CONTRIBUTING.md).

I have done so and in fact made a small addition to CONTRIBUTING.md to point out that there's also the `genDocs` nimble task.

can-lehmann:
> Actually do you have a particular feeling about `float` vs `float64`? Mostly because float implicitly is float64 on basically any reasonably modern system and I don't think I consistently used `float64` everywhere when typing things.

I don't really have a preference, though I believe I tried to use `float` in owlkettle.

owlkettle/widgets.nim
can-lehmann: I am unable to reproduce this problem.

Though the state should definitely already store the value. This is the generated code for the property hook (generated using `-d:owlkettleDebug`):

```nim
  if widget.hasValue:
    state.value = widget.valValue
  else:
    state.value = 0
  block:
    gtk_range_set_value(state.internalWidget, state.value.cdouble)
```

PhilippMDoerner: I might have just done something wrong at somewhere, if I just remove the build hook now it works for me as well.
I've thus removed said build hook, we're good.

PhilippMDoerner:
> > Actually do you have a particular feeling about `float` vs `float64`? Mostly because float implicitly is float64 on basically any reasonably modern system and I don't think I consistently used `float64` everywhere when typing things.
> 
> I don't really have a preference, though I believe I tried to use `float` in owlkettle.

In that case I swapped the rest over for consistencies sake. Also resolved the last couple issues.

PhilippMDoerner:
Updated the example to make every field manipulatable so you can essentially see the widget in all kinds of configurations.
That is except for the marks field, I can't seem to get my idea there to work properly =/

PhilippMDoerner:
Looking over AcrionRow and its inheritance chain made me realize:
ActionRow inherits from other widgets and that is modeled in owlkettle by those widgets also being there.

Would it have been better for Range to first make a `Range` renderable to then have `Scale` inherit from it?
That way anything else based on Range wouldn't need to implement property hooks etc. from Range, it could have them just by inheriting from it.

can-lehmann:
Just wanted to quickly confirm that the doc comments are **NOT** copied from the GTK docs (since GTK is licensed under the LGPL).

can-lehmann:
Looks good. I changed the example around a bit so that it additionally includes controls for the `orient`, `valuePosition` and `marks` fields (+ updated the screenshot).

PhilippMDoerner:
From what I could see you removed the fields from Base-Widget (sensitive, sizeRequest and tooltip).

Should I assume that controls on the sidebar should not include fields that a given Widget has from its Parents in general?
Or just BaseWidget (as PopOverMenu for example inherits some stuff from Popover etc.)?

can-lehmann:
I don't think I removed any fields (the diffs look much larger than they actually are, because I changed the layout to be side by side instead of top/bottom).

I think the `BaseWidget` fields were never exposed in the example: https://github.com/can-lehmann/owlkettle/pull/90/commits/d28e6c23e0af73b0f9ba55ffa7e1747c9a0dc53b

can-lehmann:
Main change was just layout, moving to adwaita + adding the controls for the remaining fields mentioned above.

PhilippMDoerner:
Ohhhh I might have added those in the #94 PR, nevermind then

can-lehmann:
Thats fine.

PR #91, opened 2023-09-30
Add docs for wrapping gtk widgets for new contributors

PhilippMDoerner:
Because I dislike having to ask the same thing twice I tend to write everything down.
So basically I had a ton of notes from wrapping the Scale widget and from our discussion surrounding that.

Given that this might be useful for any future contributors I thus then decided to clean up those notes a little and add them to the nimibook, because why not at that point.

This is the first draft on how such a chapter could look like, as per the usual all of this is freely changeable.
The point after all is to have the best introductory docs for newcomers possible to increase the likelihood of others contributing.

PR #94, opened 2023-10-02
#93 add form generator

### LINKED ISSUE #93: Add a "Form-Generating"-feature to Owlkettle
### The usecase
For all the example widgets I would like to have a generic proc or template or so that I can just hand any kind of `<WidgetName>State` and that generates me a Form where I have DataEntry fields for every field on said state (except for app and viewed as those are owlkettle internal I think?).

This would be useful to let the user see what happens under various circumstances if they manipulate the state of a widget as they please.

It could also be useful if users want to generate a form to manipulate data on the fly without having to write it by hand.

### The concept of a solution
Conceptually, I'd have assumed any approach would simply need to iterate over the fields of `<WidgetName>State` type, transform the values of the fields into an enum-type and then use fieldName and the enum-type to associate the field on  `<WidgetName>State` with a DataEntry widget. Which DataEntry widget is chosen is determined by the enum-type.

This normally wouldn't be toooooo hard without but the combination of any State type having `app` and `viewed` fields that somehow cause compiler crashes when accessed by the loop as well as the guidsl tolerating neither when statements (which would allow me to avoid those) nor macros that enable unrolling a `static: seq[string]` for-loop in the gui section makes all my typical non-macro approaches impossible.

#### Prior approach (Placeholder for like 3 other similar ones I all attempted)
Removed, it is irrelevant as I found a solution

PhilippMDoerner:
This PR adds a generic proc that generates a Box with a List of Widgets for every field in a given object.
Which mostly means for every field in a given `<Widget>State`.

This autoform was only used in the `Scale.nim` example for now, other examples shall be updated in a separate PR to keep these small and easily reviewable. 

The reasoning for why I chose a generic proc rather than writing a generic viewable I laid out in #93 

This PR actually only adds an autoform.nim file, but as this branch is one that branched off the one for  #90 it contains the same changes as that one.

This PR depends on #90 as #90 provides the scale.nim example and that was what I used as reference to implement this.

As per the usual, it's all up for change regardless. Even if I already have #95, I should be able to relatively simply adjust that to whatever we decide here as I'm semi-sure the concept of "take in a state, iterate over it and spit out a widget based on that" will remain.

closes #93


can-lehmann:
This needs to be rebased now that https://github.com/can-lehmann/owlkettle/pull/90 is merged.

PhilippMDoerner:
I'm on it.
Or rather, I'm currently revamping said form generator so that it generates a sidebar based on the changes you made to the Scale Widget.
And as part of that I'm updating accordingly

PhilippMDoerner:
Okay, this is a first draft that basically recreates the sidebar as before.
Well almost.
I don't really control the order of this one so the marks field is not quite in the middle.
Also I added a DropDown field so you can decide the ScalePosition of a given mark.

My next attempt will be to see if I can't move from a custom `toFormField` method to something that fully generates a proc like that just from a given type.

PhilippMDoerner:
@can-lehmann I've got a... let's say first draft working for auto-generating ExpanderRow-Widgets for a given object type T.
 
I'm honestly not too sure what to make of this.
I can either pretty this up, or we just leave it and instead implement custom `toFormField` procs for every custom data type we have in a seq.
Might be less of a headache overall because the generic manner of this is starting to get somewhat abstract, but on the other hand having to write your own `toFormField` for custom data types every time might be repeated effort.

Then again, I don't think this should happen *too* often.
Thoughts?

PhilippMDoerner:
I've decided to just move over to the approach where you defined a custom `toFormField` for every custom type you have.
Generating forms for every normal type in an action Row is way too annoying.

owlkettle/autoform.nim
```diff
@@ -0,0 +1,135 @@
...
+          proc changed(text: string) =
+            try:
+              state.getField(fieldName) = text.parseJson().to(typ)
+            except Exception: discard
+
+proc toFormField[T: enum](state: auto, fieldName: static string, typ: typedesc[T]): Widget =
+  let options: seq[string] = T.items.toSeq().mapIt($it)
+  return gui:
+    ComboRow:
+      title = "orient"
```
can-lehmann, line 74: fieldName

owlkettle/autoform.nim
```diff
@@ -0,0 +1,135 @@
...
+      ActionRow:
+        title = fieldName
+        Box() {.addSuffix.}:
+          Switch() {.vAlign: AlignCenter, expand: false.}:
+            state = state.getField(fieldName)
+            proc changed(newVal: bool) =
+              state.getField(fieldName) = newVal
+
+proc toFormField(state: auto, fieldName: static string, typ: typedesc[object | ref object | tuple | seq]): Widget =
+    return gui:
```
can-lehmann, line 61: Indentation

owlkettle/autoform.nim
```diff
@@ -0,0 +1,135 @@
...
+  return gui:
+    ActionRow:
+      title = fieldName
+      Entry() {.addSuffix.}:
+        text = state.getField(fieldName)
+        proc changed(text: string) =
+          state.getField(fieldName) = text
+
+proc toFormField(state: auto, fieldName: static string, typ: typedesc[bool]): Widget =
+    return gui:
```
can-lehmann, line 51: Indentation

owlkettle/autoform.nim
```diff
@@ -0,0 +1,135 @@
...
+      items = options
+      selected = ord(state.getField(fieldName))
+      proc select(enumIndex: int) =
+        state.getField(fieldName) = enumIndex.T
+
+proc toFormField(state: auto, fieldName: static string, typ: typedesc[DateTime]): Widget =
+  return gui:
+    ActionRow:
+      title = fieldName
+      Calendar {.addSuffix.}:
```
can-lehmann, line 84: Maybe open a dialog with the calendar instead?

owlkettle/autoform.nim
```diff
@@ -0,0 +1,135 @@
...
+import ./adw
+import ./widgetutils
+import ./guidsl
+import ./widgetdef
+import ./widgets
+
+macro getField*(someType: untyped, fieldName: static string): untyped =
+  nnkDotExpr.newTree(someType, ident(fieldName))
+
+proc `%`*(t: tuple): JsonNode =
```
can-lehmann, line 12: Remove, see below.

owlkettle/autoform.nim
```diff
@@ -0,0 +1,135 @@
...
+          
+          for field in fieldWidgets:
+            insert field
+
+
+proc toAutoForm*[T](app: T): Widget =
+  const ignoreFields: seq[string] = @[]
+  return toAutoForm[T](app, ignoreFields = ignoreFields)
+
+export AlignedChild
```
can-lehmann, line 135: Why is this necessary? (remove if unnecessary)

owlkettle/autoform.nim
```diff
@@ -0,0 +1,135 @@
...
+      proc select(enumIndex: int) =
+        state.getField(fieldName) = enumIndex.T
+
+proc toFormField(state: auto, fieldName: static string, typ: typedesc[DateTime]): Widget =
+  return gui:
+    ActionRow:
+      title = fieldName
+      Calendar {.addSuffix.}:
+        date = state.getField(fieldName)
+        proc daySelected(date: DateTime) =
```
can-lehmann, line 86: select

owlkettle/autoform.nim
```diff
@@ -0,0 +1,135 @@
...
+  for field, value in t.fieldPairs:
+    result.add(field, %value)
+
+## General Fields
+
+proc toFormField(state: auto, fieldName: static string, typ: typedesc[SomeFloat]): Widget =
+  return gui:
+    ActionRow:
+      title = fieldName
+      NumberEntry() {.addSuffix.}:
```
can-lehmann, line 23: Could be a FormulaEntry

owlkettle/autoform.nim
```diff
@@ -0,0 +1,135 @@
...
+    return gui:
+      ActionRow:
+        title = fieldName
+        Box() {.addSuffix.}:
+          Switch() {.vAlign: AlignCenter, expand: false.}:
+            state = state.getField(fieldName)
+            proc changed(newVal: bool) =
+              state.getField(fieldName) = newVal
+
+proc toFormField(state: auto, fieldName: static string, typ: typedesc[object | ref object | tuple | seq]): Widget =
```
can-lehmann, line 60: > I've decided to just move over to the approach where you defined a custom toFormField for every custom type you have.
> Generating forms for every normal type in an action Row is way too annoying.

This should also be removed then (?)

owlkettle/autoform.nim
PhilippMDoerner: There's a select proc for Calendar?

I'm only aware of these:
```nim
  proc daySelected(date: DateTime)
  proc nextMonth(date: DateTime)
  proc prevMonth(date: DateTime)
  proc nextYear(date: DateTime)
  proc prevYear(date: DateTime)
```
Or do you mean that this should be a dropDown widget instead to select a day?

owlkettle/autoform.nim
PhilippMDoerner: Not quite, I do need a default on what should happen for those types.
That is simply that I turn them into json and the user can update that JSON string (And those updates remain as long as the string can still be parsed into the custom type.

That way I don't *force* the user to define a `toFormField` proc for their type, it is merely an option.

If an (ugly) default like this is not desired I'll throw it out of course, was just my default approach.

owlkettle/autoform.nim
can-lehmann: There is a convenience event called `select`, see the calendar example.

owlkettle/autoform.nim
PhilippMDoerner: Done.
Basically now it shows `Date: <Stringified Date>  [Select]` with the Select button being "default" type and that opens a Custom DateDialog where you can select a date.

Looks alright?

owlkettle/autoform.nim
PhilippMDoerner: Doneso

owlkettle/autoform.nim
PhilippMDoerner: Will do (or not do) based on the outcome of our discussion around 
`proc toFormField(state: auto, fieldName: static string, typ: typedesc[object | ref object | tuple | seq]): Widget =`

owlkettle/autoform.nim
can-lehmann: The issue I have with JSON & Nim is that the mapping between the two is not bijective. There are Nim objects that cannot be mapped to JSON (e.g. cyclic references) and different types might map to the same JSON. (I am actually not exactly sure what happes for recursive types, though I would assume that it just creates an infinite recursion (?)).

From a UI perspective, the editing experience of typing JSON into an entry is also not really desirable.

There is however still the question of what to do with what to do with types that do not have a `toFormField` implementation. I think that since you want to encourage creating custom controls for your types (an appraoch I strongly agree with), it might be fine to just show some icon with a tooltip that says something like "Unable to edit [TypeName], please override toFormField.".

can-lehmann:
https://github.com/can-lehmann/owlkettle/pull/91#issuecomment-1747218057 also applies here, so I would personally prefer to merge this after the JSON discussion and then move any discussions & changes to https://github.com/can-lehmann/owlkettle/pull/95.

owlkettle/autoform.nim
PhilippMDoerner: I like the "Add a dummy-widget"-approach! I'll do that one

PhilippMDoerner:
Ohhh merge this into #95 and close this PR? Works for me

PhilippMDoerner:
Basically, if you tell me now this looks okay I'll merge this branch into #95 and close this PR

can-lehmann:
> Ohhh merge this into https://github.com/can-lehmann/owlkettle/pull/95 and close this PR? Works for me.

Sounds good.

Issue #86, opened 2023-09-17
Add `adw.StatusPage` widget


can-lehmann:
https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.StatusPage.html

can-lehmann:
Usually the docs say `since: 1.2` (see e.g. here: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.Breakpoint.html - although that is since 1.4). The CI should fail anyways if the widget is only available since 1.2, but you did not put it inside a adwaita12 block.

PhilippMDoerner:
So do we distinguish only between 1.0 and 1.2 or should the instruction be "If you see it a `since 1.X` block, put your code behind a `when defined (adwaita1X)` block?"

can-lehmann:
If a new feature is only available after 1.X, it should be in a `when defined (adwaita1X)` block, this case just did not come up yet.

PhilippMDoerner:
This likely will end up in a separate issue:
Could it make sense to refactor the adwaita bits of the library to use a flag `-d:adwaita=12` instead of `-d:adwaita12` ?

This way you could easily implement logic that allows you to compile with `-d:adwaita=14` and because 14 > 12 all when-statements requiring adwaita 1.2 will be there.

Something like:
```nim
const adwTargetVersion {.intdefine: "adwaita".}: int = 10 ## The version of adwaita 1.0

template minAdwaitaVersion(minVersion: int, body: untyped)=
  when adwTargetVersion >= minVersion:
    body
```

Useage example (compile with `-d:adwaita=14` or any value larger than 12 really):
```nim
const adwTargetVersion {.intdefine: "adwaita".}: int = 10

template minAdwaitaVersion(minVersion: int, body: untyped)=
  when adwTargetVersion >= minVersion:
    body
    
minAdwaitaVersion(12):
  echo "The adwaita version was larger than 12: ", adwTargetVersion
```

can-lehmann:
Although I think we shouldn't use a template for the comparisons, but rather just use a regular when statement. It might also be nice to split the version into a tuple: `(majorVersion, minorVersion)`.

PhilippMDoerner:
Kinda like nim does it with `NimMajor` and `NimMinor`: `AdwMajor` and `AdwMinor` defined by default as 1 and 0.

PhilippMDoerner:
I suggest moving further discussions about this to #96

[...]

Issue #85, opened 2023-09-17
Add `Expander` widget


can-lehmann:
https://docs.gtk.org/gtk4/class.Expander.html

PhilippMDoerner:
I've started working on this and would call it mostly done.

For the most part I'd wait with a PR to see the outcome of the autoForm PR in order to adjust based on whatever that outcome is.

Other than that, [my branch](https://github.com/can-lehmann/owlkettle/compare/main...PhilippMDoerner:owlkettle:%2385-add-expander) contains the following:
- Fully wrapped Widget, including bindings for the activate signal
- expander.nim example
- 200% Screenshot of expander.nim example and updated examples/README.md
- Updated widgets.md as per `nimble genDocs`

PR #95, opened 2023-10-02
Add autoform to all the widgets

PhilippMDoerner:
Basically a "continuation" of #93 . This PR simply adds autoform to all widget examples under /examples/widgets.
Again, the only files changed here are actually widgets.nim and the various example files, anything else stems from either #90 or #94.

This should only be merged (and honestly also only reviewed) after #94 was merged as it directly depends upon that PR.

The exact examples in total that autoform was added to encompass thusly:
- Scale (From PR #94 )
- Calendar
- Popover Menu
- Radio Group
- Drop Down
- Picture
- Text View

Beyond just adding autoform, I also added the fields of the BaseWidget to those examples so that users can be made more aware that they also have access to those and can see their effects on the different widgets.

Further this PR also fixes a small bug with Radio Group:
It was missing a property hook for its `orient` field. 
This PR adds the missing property hook in widgets.nim, which enables RadioGroup to change its orientation at runtime, which previously wasn't possible.


can-lehmann:
I must say that I have some concerns regarding the current UI design (even though I am very aware it is based on the Slider example). Looking through the screenshots want to be able to immediately identify what the example does. When we add a bunch of large controls at the side of each example window, this is not the case anymore.

I have not checked if this works, but I would propose moving the forms into a `Popover` that opens when clicking on a menu button in the `HeaderBar`. We should have a consistent design language for this, i.e. always using the same icon and roughly the same placement. I would propose using the "edit-symbolic" icon and putting it in the top right of the HeaderBar.

The importance of simplicity also applies to the code (for the same reasons). So the goal would be to minimize the number of lines that are affected by the `autoform` module. New users need to be able to quickly see the widget and how it is used, not have to look through a bunch of boilerplate code. I think this already looks quite good currently. You might want to consider moving all of the `toFormField` into the `autoform` module.

PhilippMDoerner:
I'm happy with just moving the `toFormField` into `autoform.nim`.

I also agree with keeping the UI simpler by not having the form visible initially but by clicking on a button. I just tend to think more along the lines of "Click button, box on the left slides in or out", kind of how Obsidian does it. See below:

[Screencast from 2023-10-04 21-09-44.webm](https://github.com/can-lehmann/owlkettle/assets/65814317/26e3b642-bdbe-4cea-822b-40cd1ad476d2)

The main issue I guess being that I'm not entirely sure how to replicate this well within gtk without overblowing the examples.
The alternative would I guess be to have 2 separate kinds of examples, a minimalistic one and a "playground" one (?) ?

can-lehmann:
We have a `Flap` widget, that does this, but as you already said, using it would complicate the example code a lot, thus violating the second requirement I formulated above.

can-lehmann:
>  The alternative would I guess be to have 2 separate kinds of examples, a minimalistic one and a "playground" one (?) ?

Interesting idea.

At some point I experimented with an idea like this:

```nim
import owlkettle
import owlkettle/playground

... Define your custom widgets ....

playground[MyWidgetState](gui(MyWidget()).build())
```

With some macros you could also do some nice things like

```nim
playground:
  MyWidget()
```

Maybe you could even support whole widget trees, so you can get controls for all widgets in the sidebar (for the `playground` stuff I would use the sidebar design of the Slider example).

```nim
playground:
  Box:
    Label:
      text = "Label 1"
    Label:
      text = "Label 2"
```

However I think that this does not conflict with the idea you looked at in this PR so far (adding controls to existing examples). The two serve quite different purposes. While `playground` is targeted at users who build larger applications and want to document/demonstrate the widgets they built, the examples are for learning owlkettle & maybe playing with the available fields.

PhilippMDoerner:
> However I think that this does not conflict with the idea you looked at in this PR so far (adding controls to existing examples

You sure that's what you meant and not "I think that this conflicts with the idea you looked at in this PR so far" or "I think that this does not work with the idea you looked at in this PR so far" ? If you did mean that, I'm getting conflicting messages on if we should proceed with this idea now or not.

Generally, I see your point there, and agree.

Regarding complexity:
Yeah I can't seem to keep them minimal enough to consider them be actually minimal.
Every example would need to contain a flap widget, a headerbar with a button and a changed proc and a call to "toAutoForm".

I'm somewhat uncertain how to proceed from here.
Throw the idea away in general? 
The idea originally was more to do an improvement for the docs once you've moved past the initial stage with owlkettle, to provide you a view into the amount of features available in a widget as currently any desire to see if a widget can do X requires you to write an example yourself.
I'm not 100% certain I find it within myself currently to turn that into a large section where you can dump in a widget and it auto generates an application with a sidebar out of it.
I'm also not 100% certain if that is actually within the scope of owlkettle or outside of it. The feature idea seems somewhat large, basically like an entire Tool for owlkettle development on its own. 
Is that too large for the scope of owlkettle? I don't know.

If we do throw autoform as a concept away, at the very least I'd want to change the Scale widget to be more minimal rather than the more complex example it currently is.

can-lehmann:
> You sure that's what you meant and not "I think that this conflicts with the idea you looked at in this PR so far" or "I think that this does not work with the idea you looked at in this PR so far" ? If you did mean that, I'm getting conflicting messages on if we should proceed with this idea now or not.

No, I truly think that there is no conflict between the two ideas, as they serve different purposes:

- This PR extends the examples to make them more interactive.
- A `playground` feature targets custom widgets the user wrote themselves.

Both would use your `autoform` implementation, but the user API would be different.

can-lehmann:
>  Is that too large for the scope of owlkettle? I don't know.

Since `autoform` is only ~150 LOC, I'd say it is within the scope of owlkettle. E.g. `dataentries` is ~250LOC and implements a mathematical expression parser, but is also part of owlkettle.

can-lehmann:
> Every example would need to contain a flap widget, a headerbar with a button and a changed proc and a call to "toAutoForm".

That is why I proposed the popover based approach. I am pretty sure it is possible to keep the required code quite small, because you only need to add a single `MenuButton` (could be generated in a proc in `autoform`).

can-lehmann:
I think I should say that I am definitely in favor of adding `autoform` to owlkettle & adding controls for all of the fields to the examples. I just think that the execution needs to be right here, as good examples are imo extremely important for learning a new library.

PhilippMDoerner:
I think the funny bit is we're both moving a mile a minute in our heads on where to go from here but I sometimes end up in completely different areas :laughing: 

Just wrote an example with the popover approach. It does work out pretty well.
The smallest I could manage to squeeze the amount of code that'd need to be added to a given example would be 2 lines:
```nim

viewable App:
  min: float = 0
  max: float = 100
  value: float = 50
  marks: seq[ScaleMark] = @[]
  inverted: bool = false
  showValue: bool = true
  stepSize: float = 5
  pageSize: float = 10
  orient: Orient = OrientX
  showFillLevel: bool = true
  precision: int64 = 1
  valuePosition: ScalePosition
  sensitive: bool = true
  tooltip: string = ""
  sizeRequest: tuple[x, y: int] = (-1, -1)

method view(app: AppState): Widget =
  result = gui:
    WindowSurface:
      defaultSize = (800, 600)
      Box(orient = OrientY):
        Box() {.expand: false.}: # Only this Box and the insert app.toAutoForm are needed in the user side, that's all.
          insert app.toAutoForm()
        
        Scale:
          min = app.min
          max = app.max
          value = app.value
          marks = app.marks
          inverted = app.inverted
          showValue = app.showValue
          stepSize = app.stepSize
          pageSize = app.pageSize
          orient = app.orient
          showFillLevel = app.showFillLevel
          precision = app.precision
          valuePosition = app.valuePosition
          sensitive = app.sensitive
          tooltip = app.tooltip
          sizeRequest = app.sizeRequest
          
          proc valueChanged(newValue: float64) =
            app.value = newValue
            echo "New value from Scale is ", $newValue

adw.brew(gui(App()))
```
[Screencast from 2023-10-04 22-34-07.webm](https://github.com/can-lehmann/owlkettle/assets/65814317/59226c20-121e-4655-9146-79e55c5661c4)

Could be 1 line if I managed to figure out how to apply `{.expand: false.}` to a widget added via `insert <Widget>`.

can-lehmann:
Looks really good! I think thats minimal enough. For the examples that already have a `HeaderBar`, you probably need to make `autoform` just emit the `MenuButton` (so that would be only a 1 line diff, which is pretty minimal imo xD).

PhilippMDoerner:
> Looks really good! I think thats minimal enough. For the examples that already have a `HeaderBar`, you probably need to make `autoform` just emit the `MenuButton` (so that would be only a 1 line diff, which is pretty minimal imo xD).

Hmmm so
`toAutoFormMenu` (emits headerbar with button) and `toAutoFormButton` (emits only the button) ?
Or do we only have `toAutoForm` which emits a menu-button and we move the Headerbar into the example-widgets (I'd insert them via direct assignment to right: `right = @[app.toAutoFormMenu()]`) ?

can-lehmann:
I think I would just use `toAutoFormMenu` and then `insert app.toAutoFormMenu() {.addRight.}`. We can also move back to `Window` (instead of `WindowSurface`).

PhilippMDoerner:
Alrighty, the recent changes should've added the new Popover version to the examples I've listed for individual widgets.
Also added the MIT License as you usually did for all the widgets.

I also did a small overhaul, mostly to "simplify"/"unify" forms for types like seq[string] etc.

If you have a field of type `MyCustomType`, you must implement  `proc toFormField(state: auto, fieldName: static string, typ: typedesc[MyCustomType]): Widget =`.

If you have a field of type `seq[MyCustomType]`, you must now implement `proc toListFormField(state: auto, fieldName: static string, index: int, typ: typedesc[MyCustomType]): Widget =`

That way you only need to worry about making a form for *one* entry, not for an entire list of them.
The way it works is that basically there's already a `toFormField` defined for `seq[T]` and that one just calls `toListFormField` to generate the individual rows, puts it all into an ExpanderRow and provides a "plus" button to add new entries.
I couldn't figure out how to easily just add a Delete Button for you, but I think that's fine.

If you have better idea for wording, shoot it out, I'll happily improve upon the meh names I landed on for these procs. Naming is hard.

Now with that out of the way, I have the following questions:
1) Should the form and the entire concept remain being called "autoform"? It was literally just the first name that popped, another one of those "If you got a better name, feel free to suggest one, I'll be happy to rename module procs and docs accordingly"
2) I noticed that while going over the widgets: Not every widget has an example like this, right? Immediate things would be stuff like "Button" or "Label". Would it make sense to add more widget examples? If yes, should I add them in this PR or should we first merge this and *then* I make a new PR with more widget example applications ? (The goal being 1 example app per widget in widgets.nim and adw.nim)

owlkettle/widgets.nim
```diff
@@ -1693,6 +1693,10 @@ renderable RadioGroup of BaseWidget:
       for row in state.rows:
         gtk_box_set_spacing(row.box, state.rowSpacing.cint)
   
+  hooks orient:
+    property:
+      gtk_orientable_set_orientation(state.internalWidget, state.orient.toGtk())
```
PhilippMDoerner, line 1698: This one is actually a bug-fix: Radio-buttons did not react to if you changed the orientation at runtime because they had no property hook

can-lehmann: You are right, that seems to be missing :+1:

can-lehmann:
> Should the form and the entire concept remain being called "autoform"? It was literally just the first name that popped, another one of those "If you got a better name, feel free to suggest one, I'll be happy to rename module procs and docs accordingly"

We could already call it `playground` in preparation for potential future applications.

> I noticed that while going over the widgets: Not every widget has an example like this, right? Immediate things would be stuff like "Button" or "Label". Would it make sense to add more widget examples? If yes, should I add them in this PR or should we first merge this and then I make a new PR with more widget example applications ? (The goal being 1 example app per widget in widgets.nim and adw.nim)

I decided to only create examples for the more complex widgets pretty early on, since I trust that users should be able to figure out how to use the basic widgets like `Button`, `Label`, ... from the docs, tutorial and other examples. I don't think that the added maintenance work for examples for every widget is worth it.

can-lehmann:
I played with some of the examples and I very much like it! I think that the new design works quite well.

A few general notes:

- Currently all examples have two header bars, you need to add the `HeaderBar`s using `addTitlebar`.
- All buttons within a `HeaderBar` should have the style `ButtonFlat` (as in basically all other Adwaita apps, probably also specified in the HIG)
- I would show the currently selected date for the `DateTime` form in the `description` of the `ActionRow`, which makes it take up less space horizontally.
- There is lots of code duplication between `toListFormField` and `toFormField`, this should be reduced (it might be possible to just remove the difference between the two procs entirely?).

examples/README.md
```diff
@@ -149,23 +149,23 @@ The `widgets` directory contains examples for how to use different widgets.
...
+    <td><img alt="Radio Group" src="../docs/assets/examples/radio_group.png" width="841px"></td>
   </tr>
   <tr>
     <td><a href="https://github.com/can-lehmann/owlkettle/blob/main/examples/widgets/scale.nim">Scale</a></td>
     <td><img alt="Scale Widget" src="../docs/assets/examples/scale.png" width="922px"></td>
   </tr>
   <tr>
     <td><a href="https://github.com/can-lehmann/owlkettle/blob/main/examples/widgets/text_view.nim">Text View</a></td>
-    <td><img alt="Text View" src="../docs/assets/examples/text_view.png" width="757px"></td>
+    <td><img alt="Text View" src="../docs/assets/examples/text_view.png" width="608px"></td>
```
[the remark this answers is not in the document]
PhilippMDoerner: Done

PhilippMDoerner:
Punkt:

1) Done
2) To be done
3) Done
4) That one we run into a conceptual issue.

Particularly for 4: The **only** reason these are separate is because in `toFormField` i need to generate expressions along the lines of `state.myField`, in `toListFormField` I need to generate expressions along the lines of `state.myField[index]`.
That is **the only** reason these two functions are separate.

I want very much to act directly on state because I know that is a ref that is shared between the main widget in the example as well as the every proc autoform that gets passed said state.
I can not assume that every field on autoform is a ref type and that I don't get copies somewhere that'd lead to the autoform widgets modifying their own copies of a values on the `state` without modifying the actual thing on `state`.

I'll launch an attempt at using `var fieldValue: <Type of fieldvalue>` instead of `state: auto, fieldName: static string, typ: <Type of fieldValue>` and `state: auto, fieldName: static string, typ: <Type of fieldvalue>` in the hopes that this'll get nim to use references under the hood. However, I make no guarantees and no promises.

PhilippMDoerner:
I can't seem to find any approach that allows me to unify `toFormField` and `toListFormField` and get rid of the code duplication.
 
The approach of "just pass the fieldValue as a mutable parameter `var T` doesn't work because you can't capture it in an event-handler:
```nim
proc toFormField[T: SomeNumber](fieldName: static string, fieldValue: var T): Widget =
  ## Provides a form field for any number type
  return gui:
    ActionRow:
      title = fieldName
      FormulaEntry() {.addSuffix.}:
        value = fieldValue.float
        xAlign = 1.0
        maxWidth = 8
        proc changed(value: float) =
          fieldValue = value.T
```
```
'fieldValue' is of type <var float> which cannot be captured as it would violate memory safety, declared here: /home/philipp/dev/owlkettle/owlkettle/autoform.nim(35, 33); using '-d:nimNoLentIterators' helps in some cases. Consider using a <ref var float> which can be captured.
```
I also did various attempts centered around templates, the most recent one being to turn the gui block into a template like this:
```nim
template stringField(fieldName: string, fieldExpression: untyped): Widget =
  ## Provides a form field for string
  gui:
    ActionRow:
      title = fieldName & "I am from the trees"
      Entry() {.addSuffix.}:
        text = fieldExpression
        proc changed(text: string) =
          fieldExpression = text

proc toFormField(state: auto, fieldName: static string, typ: typedesc[string]): Widget =
  ## Provides a form field for string
  return stringField(fieldName):
    state.getField(fieldName)

proc toListFormField(state: auto, fieldName: static string, index: int, typ: typedesc[string]): Widget =
  ## Provides a form to display a single entry of type `string` in a list of `string` entries.
  return stringField(fieldName & $index):
    state.getField(fieldName)[index]
```
All of them blow up with the error:
```nim
stack trace: (most recent call last)
/home/philipp/dev/owlkettle/owlkettle/guidsl.nim(293, 17) gui
/home/philipp/dev/owlkettle/owlkettle/guidsl.nim(111, 36) parseGui
/home/philipp/dev/owlkettle/owlkettle/guidsl.nim(94, 11) parseGui
/home/philipp/dev/owlkettle/owlkettle/guidsl.nim(111, 36) parseGui
/home/philipp/dev/owlkettle/owlkettle/guidsl.nim(94, 11) parseGui
/home/philipp/dev/owlkettle/owlkettle/guidsl.nim(111, 36) parseGui
/home/philipp/dev/owlkettle/owlkettle/guidsl.nim(113, 11) parseGui
/home/philipp/.choosenim/toolchains/nim-2.0.0/lib/std/assertions.nim(41, 14) failedAssertImpl
/home/philipp/.choosenim/toolchains/nim-2.0.0/lib/std/assertions.nim(36, 13) raiseAssert
/home/philipp/.choosenim/toolchains/nim-2.0.0/lib/system/fatal.nim(53, 5) sysFatal
/home/philipp/dev/owlkettle/examples/widgets/drop_down.nim(38, 12) template/generic instantiation of `gui` from here
/home/philipp/dev/owlkettle/examples/widgets/drop_down.nim(46, 19) template/generic instantiation of `toAutoFormMenu` from here
/home/philipp/dev/owlkettle/owlkettle/autoform.nim(321, 27) template/generic instantiation of `toAutoFormMenu` from here
/home/philipp/dev/owlkettle/owlkettle/autoform.nim(302, 28) template/generic instantiation of `toFormField` from here
/home/philipp/dev/owlkettle/owlkettle/autoform.nim(277, 38) template/generic instantiation of `toListFormField` from here
/home/philipp/dev/owlkettle/owlkettle/autoform.nim(184, 21) template/generic instantiation of `stringField` from here
/home/philipp/.choosenim/toolchains/nim-2.0.0/lib/system/fatal.nim(53, 5) Error: unhandled exception: /home/philipp/dev/owlkettle/owlkettle/guidsl.nim(113, 7) `node[0].isName`  [AssertionDefect]
```

Beef had some approach regarding turning `toFormField` procs in general into templates, but that ran into the same error.
He mentioned something about the gui macro being an untyped macro causing some limitations there (?).

Over all, I got nothing regarding deduplicating this.
That is, outside of using `addr` to create pointers to the fields on `AppState` (even if they're value-types), which is an approach so cursed that I explored whether it may be feasible in theory but did not attempt.

PhilippMDoerner:
> All buttons within a HeaderBar should have the style ButtonFlat (as in basically all other Adwaita apps, probably also specified in the HIG)

Updated this just now:
- Changed the autoform menubutton to be flat
- Changed the menubutton for popover menu and *almost* all the headerbar button in the picture example into Flat buttons. I did not do that for the "open" button on picture because that overwrites its "ButtonSuggested" styling and I assumed that was important.

examples/widgets/popover_menu.nim
```diff
@@ -21,20 +21,32 @@
...
 method view(app: AppState): Widget =
   result = gui:
     Window:
-      title = "Popover Menu Demo"
-      defaultSize = (300, 200)
-      
+      defaultSize = (800, 600)
+
       HeaderBar {.addTitlebar.}:
+        insert(app.toAutoFormMenu(ignoreFields = @["pixbuf", "loading"], sizeRequest = (400, 300))) {.addRight.}
```
can-lehmann, line 38: ignoreFields is not necessary, is sizeRequest necessary?

examples/widgets/popover_menu.nim
```diff
@@ -21,20 +21,32 @@
...
+  offset: tuple[x, y: int] = (0, 0)
+  position: PopoverPosition = PopoverBottom
 
 method view(app: AppState): Widget =
   result = gui:
     Window:
-      title = "Popover Menu Demo"
-      defaultSize = (300, 200)
-      
+      defaultSize = (800, 600)
```
can-lehmann, line 35: Revert title & default size

examples/widgets/radio_group.nim
```diff
@@ -20,73 +20,52 @@
...
-        Button {.addRight.}:
-          icon = "go-down-symbolic"
-          tooltip = "Select Next"
-          style = [ButtonFlat]
-          
-          proc clicked() =
-            app.selected += 1
+        insert(app.toAutoFormMenu(ignoreFields = @["pixbuf", "loading"], sizeRequest = (500, 520))) {.addRight.}
+
+      Box:
```
can-lehmann, line 49: ScrolledWindow, see original version of example

examples/widgets/radio_group.nim
```diff
@@ -20,73 +20,52 @@
...
+      defaultSize = (800, 600)
       HeaderBar {.addTitlebar.}:
         WindowTitle {.addTitle.}:
           title = "Radio Group Example"
           if app.selected in 0..<app.options.len:
             subtitle = app.options[app.selected]
           else:
             subtitle = "Invalid Item"
-        
-        Button {.addLeft.}:
```
can-lehmann, line 42: Revert removing the buttons.

examples/widgets/scale.nim
```diff
@@ -36,163 +36,37 @@ viewable App:
...
+  tooltip: string = ""
+  sizeRequest: tuple[x, y: int] = (-1, -1)
 
 method view(app: AppState): Widget =
   result = gui:
-    WindowSurface:
+    Window():
       defaultSize = (800, 600)
+      HeaderBar() {.addTitlebar.}:
+        Label(text = "Scale Example") {.addTitle.}
```
can-lehmann, line 48: Use `Window.title`

examples/widgets/text_view.nim
```diff
@@ -22,22 +22,32 @@
...
+  acceptsTab: bool = true
+  indent: int = 0
+  sensitive: bool = true
+  sizeRequest: tuple[x, y: int] = (-1, -1) 
+  tooltip: string = "" 
 
 method view(app: AppState): Widget =
   result = gui:
     Window:
-      title = "Text View Example"
```
can-lehmann, line 32: Add back the `Window.title`

owlkettle/playground.nim
```diff
@@ -0,0 +1,316 @@
...
+        res = DialogAccept
+        
+      DialogButton {.addButton.}:
+        text = "Cancel"
+        res = DialogCancel
+        
+      Box(orient = OrientY):
+        Calendar:
+          date = state.date
+          proc daySelected(date: DateTime) =
```
can-lehmann, line 123: Use select callback

can-lehmann:
> That is, outside of using addr to create pointers to the fields on AppState (even if they're value-types), which is an approach so cursed that I explored whether it may be feasible in theory but did not attempt.

That is actually the way to go here from what I can see. However, since this PR is currently blocking your other widgets, I would be fine with merging this as is (provided the things in the review above are fixed) and going over `playground`/`autoform` again at a later stage & refactoring it to use `ptr`.

examples/widgets/popover_menu.nim
PhilippMDoerner: I'd like to keep the sizeRequest as the default size of the menu is more suited to an amount of fields like Scale has.
Thus (in my opinion) it looks nicer if the height is just enough to cover the fields.
With the default values it looks like this:
image

examples/widgets/popover_menu.nim
PhilippMDoerner: Done, though I changed the width to 400. With the additional menubutton in the titlebar it otherwise lead to the title of the window not being fully shown.
image

For comparison 400:
image

examples/widgets/radio_group.nim
PhilippMDoerner: Oh? I thought them superfluous since they were basically doing the exact same thing as autoform/playground was doing - Manipulating the options for demonstrative purposes.

PhilippMDoerner:
I addressed the remaining points and I'm happy to tackle the refactoring PR with addr as the next one.
Or rather I can combine those, as you mentioned preferring larger PRs? 
Make a combo PR for both #85, #86 and add the playground refactor on top of that?

can-lehmann:
I must say that I am really happy with this PR at this point. Good work :smile: !

> I addressed the remaining points and I'm happy to tackle the refactoring PR with addr as the next one.
Or rather I can combine those, as you mentioned preferring larger PRs?
Make a combo PR for both https://github.com/can-lehmann/owlkettle/issues/85, https://github.com/can-lehmann/owlkettle/issues/86 and add the playground refactor on top of that?

The point about larger PRs was mainly about when they actively influence each other. E.g every time you changed something in  #94, this impacted this PR, so thats why I prefer one single, larger PR in that case. If the features are independent, having separate PRs is of course nicer, however, I would also be fine with a single larger PR if that is less work for you. The main issue I have is just when the features actively depend on each other, which makes reviewing quite difficult when they are spread across multiple PRs.

If I remember correctly, you already implemented https://github.com/can-lehmann/owlkettle/issues/85 and https://github.com/can-lehmann/owlkettle/issues/86 in large parts. Given that the (at least planned) release date for 3.0.0 approaches quickly, it would make sense to merge them before then.

can-lehmann:
Looks like the already paid off. While playing with the examples, I found a pretty stupid bug: https://github.com/can-lehmann/owlkettle/commit/8e915e3a26a92540c5628d04a737b7e93227dde7

PR #97, opened 2023-10-03
#96 adwaita versions

### LINKED ISSUE #96: Allow specifying arbitrary adwaita versions
Issue written based on the discussion in #86 

Currently owlkettle requires you to add the flag `-d:adwaita12` if you wish to compile for environments where you know the adwaita lib is installed with a version of 1.2 or higher.

This is done for widgets that were added in version 1.2 of the adwaita lib.

This doesn't scale well if later down the road we have widgets that were added in 1.4, introducing a new `-d:adwaita14` flag and now you need to specify both `-d:adwaita12` and `-d:adwaita14` if you need both their widgets.

Thus it makes sense to rather move over to compilation flags such as `-d:adwMajor=1 -d=adwMinor=2` where you define a specific version and based on that sections of code are either compiled in or out.

PhilippMDoerner:
A mechanism to specify various adwaita versions and gtk version with just 3 different flags:
-d:adwminor
-d:adwmajor
-d:gtkminor

There is no -d:gtkmajor because that needs to be 4 as this lib was written specifically for 4.

Further details as discussed in #96

Closes #96

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -41,6 +41,7 @@ jobs:
         run: |
           mkdir _site
           cp -r compiledBook/* _site
+          cd _site && ls -l
       - name: Upload site files
         uses: actions/upload-pages-artifact@v1 # This will automatically upload an artifact from the '/_site' directory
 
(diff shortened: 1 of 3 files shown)


can-lehmann:
It should be `AdwVersion >= (1, 2)`, shouldn't it?

PhilippMDoerner:
You are correct, really dumb slip-up.
Fixed it

PR #109, opened 2023-10-10
Add search entry

PhilippMDoerner:
And here we go, yet another widget.

This one is not fully supported in its feature scope as the property [Gtk.SearchEntry:activates-default](https://docs.gtk.org/gtk4/property.SearchEntry.activates-default.html) is not wrapped as discussed.

However, it still serves as a nice text entry with support for some hotkeys to trigger various search signals.
The one thing I'm confused by is search-started though.

I did wrap it, however I have completely failed at figuring out how to trigger it. :man_shrugging: 
Should I remove it because of that or leave it in?


owlkettle/widgets.nim
```diff
@@ -2012,6 +2012,59 @@ renderable ModelButton of BaseWidget:
             proc clicked() =
               echo "Clicked " & $it
 
+renderable SearchEntry of BaseWidget:
+  # child: GtkWidget # This is currently not supported
+  searchDelay: uint = 100 ## Determines the minimum time after a `searchChanged` event occurred before the next can be emitted. Only available when compiling for gtk 4.8
```
PhilippMDoerner, line 2017: Memo to me, still requires a text field so we can have 2 way bindings

owlkettle/widgets.nim
```diff
@@ -2012,6 +2012,66 @@ renderable ModelButton of BaseWidget:
...
+  hooks:
+    beforeBuild:
+      state.internalWidget = gtk_search_entry_new()
+    connectEvents:
+      proc searchCallback(widget: GtkWidget, data: ptr EventObj[proc(searchString: string)]) =
+        let searchString = $gtk_editable_get_text(widget)
+        data[].callback(searchString)
+        data[].redraw()
+      
+      proc changedCallback(widget: GtkWidget, data: ptr EventObj[proc(searchString: string)]) =
```
PhilippMDoerner, line 2037: Note: I separated out the changedCallback from the searchCallback.
Only the changedCallback gets to update the State.
I'm not sure if that's okay or not okay, but it seemed right to me as I would always assume the last value that `changed` returns to me to be the value that is written in text.

PhilippMDoerner:
This blew up due to exceeding the rate limit: `"message":"API rate limit exceeded for ...`
I think triggering these again should do the trick

examples/widgets/search_entry.nim
```diff
@@ -0,0 +1,58 @@
...
+          sizeRequest = app.sizeRequest
+          
+          proc activate(searchString: string) =
+            echo "Search triggered: ", searchString
+          
+          proc nextMatch() = echo "Next Match"
+          proc previousMatch() = echo "Prior Match"
+          proc searchStarted(searchString: string) = echo "Search Started: ", searchString
+          proc changed(searchString: string) = echo "Search Changed: ", searchString
+          proc stopSearch(searchString: string) = echo "Search Stopped: ", searchString
```
can-lehmann, line 56: I would put an actual searchable list in this example (just a `ListBox` in a `ScrollableWindow` with ~100 autogenerated items), so that it demonstrates what the nextMatch, previousMatch, ... callbacks should do exactly. Also not sure what the HIG says about placement of search entries, but I am under the impression that putting it in the Header bar might be nicer in that case.

owlkettle/widgets.nim
```diff
@@ -2012,6 +2012,66 @@ renderable ModelButton of BaseWidget:
...
+      state.connect(state.searchStarted, "search-changed", searchCallback)
+      state.connect(state.stopSearch, "stop-search", searchCallback)
+    disconnectEvents:
+      state.internalWidget.disconnect(state.activate)
+      state.internalWidget.disconnect(state.nextMatch)
+      state.internalWidget.disconnect(state.previousMatch)
+      state.internalWidget.disconnect(state.changed)
+      state.internalWidget.disconnect(state.searchStarted)
+      state.internalWidget.disconnect(state.stopSearch)
+
```
can-lehmann, line 2056: Text hook is missing. See other PR.

examples/widgets/search_entry.nim
PhilippMDoerner: I did tackle this, not sure if that is to your satisfaction though.
Introduced a couple more fields to the example (selected, items and filteredItems), triggered filtering of the items as you type in the searchbar, I only render the filtered items and mark the one that is selected as bold.
With the nextMatch/previousMatch you can then shift the index around and by hitting enter it prints out what you currently have selected.

PhilippMDoerner:
I did some more digging into capture:
I think the entire `capture` thing is something owlkettle *can not* support in general.
At least that's the vibes for some common usecases for it that I'm getting, which might be circular references.

Gumb referred me to the rust bindings:
https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.SearchBar.html
> For keyboard presses to start a search, the search bar must be told of a widget to capture key events from through [set_key_capture_widget()](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.SearchBar.html#method.set_key_capture_widget). This widget will typically be the top-level window, or a parent container of the search bar. Common shortcuts such as Ctrl+F should be handled as an application action, or through the menu items.

That smells like requiring circular references to me, where e.g. `Window` contains references to all of its children, which have refs to their children, one of which eventually is `SearchEntry` which then needs to contain a reference to `Window`.

The problem I see is that I don't think we *can* build such circular references (?). At least I don't know any syntax that'd allow me to figure out how I'd fetch a ref off of something like window or the containing Box or sth.

Any thoughts?

examples/widgets/search_entry.nim
PhilippMDoerner: Fixed the spacing issue, I was silly and forgot an expand: false for the label currently being the "target"

owlkettle/widgets.nim
can-lehmann: If the last value of `changed` is indeed always the value currently in the `SearchEntry` that is good. Although I would remove the `searchString` parameter from all other events in order to avoid confusion (& force the user to do a proper 2-way binding). Is there anything that prevents you from updating the state in the other events though?

can-lehmann:
> Any thoughts?

From what I can see, you would indeed need refs, which are currently not implemented. So it is fine to not wrap that field.

examples/widgets/search_entry.nim
```diff
@@ -0,0 +1,88 @@
...
+            app.selected = 0
+            app.filteredItems = app.items.filterIt(searchString in it)
+          
+          proc stopSearch(searchString: string) = 
+            echo "Search Stopped"
+            app.text = ""
+            app.filteredItems = app.items
+          
+      ScrolledWindow:
+        Box(orient = OrientY):
```
can-lehmann, line 80: This should be a `ListBox` which allows for selection (2 way bind `app.selected` to `ListBox.selected`). The labels probably look nicer when left aligned + ~6 pixel margin. In case of the ListBox I would also remove the `<b>` for the selected item and use ListBox.selected instead.

can-lehmann:
>  The one thing I'm confused by is search-started though.
> I did wrap it, however I have completely failed at figuring out how to trigger it. 🤷‍♂️
Should I remove it because of that or leave it in?

I assume it is only relevant when used in combination with a key capture widget or a search bar, as it probably triggers when the user types into the key capture widget and the search entry should appear.

owlkettle/widgets.nim
PhilippMDoerner: Not really and honestly I don't think it makes much of a difference. I can do both, you got a preference (removing search-string from the other callbacks vs. every callback gets to update the `text` field) ?

examples/widgets/search_entry.nim
```diff
@@ -0,0 +1,88 @@
...
+viewable App:
+  searchDelay: uint = 100
+  text: string = ""
+  placeholderText: string = "Search List"
+  sensitive: bool = true
+  tooltip: string = ""
+  sizeRequest: tuple[x, y: int] = (-1, -1)
+  
+  items: seq[string] = mapIt(0..<100, "Search Entry " & $it)
+  filteredItems: seq[string] = mapIt(0..<100, "Search Entry " & $it)
```
can-lehmann, line 35:
```suggestion
  items: seq[string] = mapIt(0..<100, "Item " & $it)
  filteredItems: seq[string] = mapIt(0..<100, "Item " & $it)
```

Otherwise `Search Entry` refers to two different things (1. a search result 2. the `SearchEntry`).

owlkettle/widgets.nim
can-lehmann: I think after playing with the example it makes more sense to only update in `changed`.

PhilippMDoerner: Done

examples/widgets/search_entry.nim
PhilippMDoerner: Done, it's listBox and I used a Box to be able to align left since I can't set hAlign within an adder to ListBox (ListBox doesn't take in seq[BoxChild], but rather seq[Widget]).
```nim
      ScrolledWindow:
        ListBox:
          selected = [app.selected].toHashSet()
          selectionMode = SelectionSingle
          
          proc select(rows: Hashset[int]) =
            for num in rows:
              app.selected = num
          
          for index, item in app.filteredItems:
            Box():
              Label(text = item, margin = 6) {.hAlign: AlignStart, expand: false.}
        
```

PhilippMDoerner:
Alright, feedback resolved so far

can-lehmann:
LGTM!

PR #110, opened 2023-10-10
Add password entry

PhilippMDoerner:
Adds password entry.

This is not the full password entry wrapped. It also has a field `menuModel` that somehow interacts with GMenu's.
I started added bindings for those, but given that that is a larger task I'd like to outsource wrapping of GMenu into a different PR and do that at a different time when I'm a bit more mentally present.

On its own this widget is still useful, as you can enter stuff into it, hit enter and that triggers a callback. So yay, login screens or sth.


PhilippMDoerner:
Okay not sure what this error is about, all the examples run and compile locally on my end =/

PhilippMDoerner:
I don't understand.

The error is in my property hook:
```nim
  hooks activatesDefault:
    property:
      let value = g_value_new(state.activatesDefault)
      g_object_set_property(state.internalWidget.pointer, "activates-default", value.addr)
      g_value_unset(value.addr)
 ```
The `value.addr` blows up on 1.6.12 (and I assume 1.6. generally).

Meanwhile there's other hooks that do the **exact same thing**, e.g. ModelButton and its text property:

```nim  
  hooks text:
    property:
      var value = g_value_new(state.text)
      g_object_set_property(state.internalWidget.pointer, "text", value.addr)
      g_value_unset(value.addr)
```
I don't quite get what makes mine so different that it triggers compiler errors.

can-lehmann:
The error says `owlkettle/widgets.nim(3651, 88) Error: expression has no address; maybe use 'unsafeAddr'`, so you need to use `var`.

PhilippMDoerner:
The var vs let change completely whiffed past me. Thanks! Yep that fixed it.

owlkettle/widgets.nim
```diff
@@ -3616,7 +3616,48 @@ renderable Expander of BaseWidget:
...
-    
+
+renderable PasswordEntry of BaseWidget:
+  # menuModel: GtkMenuModel # Not yet supported
+  activatesDefault: bool = true
+  placeholderText: string = "Password"
+  showPeekIcon: bool = true
+  
+  proc activate(password: string) ## Triggered when the user "activated" the entry e.g. by hitting "enter" key while PasswordEntry is in focus. 
+
```
can-lehmann, line 3627: This needs at least a `text` field and a `changed` proc similar to `Entry`.

PhilippMDoerner: Done

owlkettle/widgets.nim
```diff
@@ -3616,7 +3616,53 @@ renderable Expander of BaseWidget:
...
+        PasswordEntryState(data[].widget).text = password
+        data[].callback(password)
+        data[].redraw()
+
+      state.connect(state.activate, "activate", activateEventCallback)
+      state.connect(state.changed, "changed", activateEventCallback)
+    disconnectEvents:
+      disconnect(state.internalWidget, state.activate)
+      disconnect(state.internalWidget, state.changed)
+
```
can-lehmann, line 3648: hook for `text` seems to be missing. It should be the same as in `Entry`. (Potential refactor: introduce an `Editable` viewable that handles `text` & `changed`).

examples/widgets/password_entry.nim
```diff
@@ -0,0 +1,57 @@
...
+          showPeekIcon = app.showPeekIcon
+          sensitive = app.sensitive
+          tooltip = app.tooltip
+          sizeRequest = app.sizeRequest
+        
+          proc activate(pw: string) =
+            echo "New Password: ", pw
+            
+          proc changed(pw: string) =
+            echo "Change in Password: ", pw
```
can-lehmann, line 56: It would be nice to demonstrate the 2-way binding in the example. (e.g. clear text on activate & show password length in window subtitle).

owlkettle/widgets.nim
PhilippMDoerner: I'll add that refactor to the list of refactors in the discussion to be done in a separate PR

owlkettle/widgets.nim
```diff
@@ -3836,7 +3836,59 @@ renderable Expander of BaseWidget:
...
-    
+
+renderable PasswordEntry of BaseWidget:
+  text: string
+  # menuModel: GtkMenuModel # Not yet supported
+  activatesDefault: bool = true
+  placeholderText: string = "Password"
+  showPeekIcon: bool = true
+  
+  proc activate(password: string) ## Triggered when the user "activated" the entry e.g. by hitting "enter" key while PasswordEntry is in focus.
```
can-lehmann, line 3847: The convention so far is to not pass the text in `activate`

PhilippMDoerner: Done

can-lehmann:
Looks good.

PR #113, opened 2023-10-14
Add action bar

PhilippMDoerner:
Adds the action bar widget + example

Also I just noticed that Centerbox also was an orientable and completely forgot to add that behaviour to centerbox.

The first commit of this PR fixes that.

Note: The main reason the example doesn't use a Box is that it doesn't appear to animate properly if I do it that way, making the example overall less impressive and "smooth".
So I opted to have the example display one "Dataset" that the user can "delete" (and reset to the old value if they do).
I think that's enough to get the usefulness across (?).


can-lehmann:
Looks good!

PR #114, opened 2023-10-14
Refactor bindings into a bindings directory and move adw.nim bindings into their own module

PhilippMDoerner:
Based on [this comment](https://github.com/can-lehmann/owlkettle/discussions/106#discussioncomment-7243196) from #106.

First stab at a PR for completing one of the desired general refactors, in this case moving bindings to a bindings directory and moving the adwaita bindings into their own module.

I'd keep this as a draft until the other 5 PRs are merged, because I want to make doubly sure that scenarios where this branch gets merged, then one of the others and suddenly there's a gtk.nim again outside of the bindings folder don't occur..
Then I'll merge main back into this branch, deal with the rest of the refactors that'll arise and then we could merge this PR.

That is assuming that there's no links to specific lines in the adw.nim module or direct links to the gtk.nim file (?)
I'm at least no aware of any.


owlkettle/bindings/adw.nim
```diff
@@ -0,0 +1,141 @@
+import ./gtk
```
can-lehmann, line 1: This file is missing license header

PhilippMDoerner: Done

can-lehmann:
Looks good!

PR #116, opened 2023-10-21
Add adw headerbar + support for decorationLayout to GtkHeaderbar

PhilippMDoerner:


owlkettle/adw.nim
```diff
@@ -563,6 +565,119 @@ proc `valSwipe=`*(flap: Flap, swipe: bool) =
...
+      state.updateChild(
+        state.titleWidget,
+        widget.valTitleWidget,
+        adw_header_bar_set_title_widget
+      )
+  
+  hooks showBackButton:
+    property:
+      when AdwVersion >= (1, 4):
+        echo "Run"
```
can-lehmann, line 654: Remove `echo` call

owlkettle/adw.nim
```diff
@@ -563,6 +565,119 @@ proc `valSwipe=`*(flap: Flap, swipe: bool) =
...
+  
+  adder addStart:
+    ## Adds a widget to the left side of the HeaderBar.
+    widget.hasPackStart = true
+    widget.valPackStart.add(child)
+  
+  adder addEnd:
+    ## Adds a widget to the right side of the HeaderBar.
+    widget.hasPackEnd = true
+    widget.valPackEnd.add(child)
```
can-lehmann, line 670: Naming should be consistent with `HeaderBar`

owlkettle/adw.nim
```diff
@@ -563,6 +565,119 @@ proc `valSwipe=`*(flap: Flap, swipe: bool) =
...
+  WindowControlIcon = "icon"
+  WindowControlMenu = "menu"
+
+proc toLayoutString(buttons: seq[WindowControlButton]): string =
+  return buttons.mapIt($it).join(",")
+
+renderable AdwHeaderBar of BaseWidget:
+  ## Adwaita Headerbar that combines GTK Headerbar and WindowControls.
+  packEnd: seq[Widget]
+  packStart: seq[Widget]
```
can-lehmann, line 584: Naming should be consistent with `HeaderBar`

owlkettle/adw.nim
```diff
@@ -563,6 +565,119 @@ proc `valSwipe=`*(flap: Flap, swipe: bool) =
...
+  WindowControlMinimize = "minimize"
+  WindowControlMaximize = "maximize"
+  WindowControlClose = "close"
+  WindowControlIcon = "icon"
+  WindowControlMenu = "menu"
+
+proc toLayoutString(buttons: seq[WindowControlButton]): string =
+  return buttons.mapIt($it).join(",")
+
+renderable AdwHeaderBar of BaseWidget:
```
can-lehmann, line 581: We need a consistent way of dealing with name collisions. I think the ideal solution would be to support qualified names in `guidsl`. This is a reminder to myself to implement that.

owlkettle/adw.nim
```diff
@@ -563,6 +565,119 @@ proc `valSwipe=`*(flap: Flap, swipe: bool) =
   flap.valSwipeToOpen = swipe
   flap.valSwipeToClose = swipe
 
+type PackType* = enum
+  PackStart, PackEnd
```
can-lehmann, line 569: Not used?

PhilippMDoerner: Huh, they'll plop up again in a different PR, could've sworn I saw a necessity for them here as well. Oh well, removed here.

owlkettle/adw.nim
PhilippMDoerner: Honestly I like the Adw Prefix simply because it informs me where the widget comes from.
It also tells me that if I have the chance to choose between 2 widget versions, I should use the adwaita one since its likely to have more features.

can-lehmann: I added support for qualified names, however after prototyping a bit, I think I agree with you that a prefix is the better solution here.

PhilippMDoerner: For consistencies sake then the following question:
I assume that we want to use the Adw prefix for every widget where we have ambiguous naming (or maybe add the prefix for all widgets in adw?).
Should we rename WindowSurface to AdwWindow for naming consistency, given that that's a breaking change and we have 3.0 just around the corner?

can-lehmann: It would make sense, yes.

PhilippMDoerner:
Alrighty, I updated this PR, it now adds both support for `decorationLayout` to `HeaderBar` ( #123)  as well as introducing `AdwHeaderBar`.
This was done as `AdwHeaderBar` also has decorationLayout and all the bells and whistles that come with that.
This allowed me some refactoring, where I managed to remove the doubly-added `WindowControls` enum (as both the PR for WindowControls and this PR defined it) as well as their `toLayoutString` procs.

Further I unified how their hooks work, so AdwHeaderBar now **also** has `Option[string]` for decorationLayout and provides 2 setters of `DecorationLayout` and `Option[DecorationLayout]`, all using the same types etc.

examples/widgets/adw/header_bar.nim
```diff
@@ -0,0 +1,71 @@
...
+          proc clicked() = echo "Clicked 1"
+          
+        Button(text = "2") {.addLeft.}:
+          style = [ButtonFlat]
+          proc clicked() = echo "Clicked 2"
+        
+        
+        Box() {.addTitle.}:
+          Label(text = "Title Widget")
+          Icon(name="go-home-symbolic") {.expand: false.}
```
can-lehmann, line 70: I would add a `if AdwVersion >= (1, 4):` here, so that the example also compiles without adwaita 1.4.

can-lehmann (changes requested):

can-lehmann:
Ah sorry, I almost forgot (again): Since this is a breaking change, it should be documented in https://github.com/can-lehmann/owlkettle/blob/main/docs/migrating_2_to_3.md so that we have a proper migration guide.

PhilippMDoerner:
Do we need to add some for the change of compiler flag as well?

can-lehmann:
> Do we need to add some for the change of compiler flags for adw and gtk as well?
Edit: I'll just assume yes since I would like to receive such instructions if I had an owlkettle application.

Good point, makes sense to also add that.

can-lehmann (approved):
Looks good!

PR #117, opened 2023-10-21
Add banner

PhilippMDoerner:


can-lehmann:
CI fails because the example is compiled without adwaita 1.3. I think we currently don't have a proper solution for this. Maybe just exclude the example from `nimble examples` if the adwaita version flag is not set as required?

PhilippMDoerner:
Shouldn't we be able to just compile for adw? I mean, libadwaita is present in the install, though I don't know if 22.04 contains libadwaita 1.4 I would assume so (?)

can-lehmann:
At least according to https://packages.ubuntu.com/jammy/libadwaita-1-0, its only version 1.1.0 (although the package is confusingly named `1-0`)

PhilippMDoerner:
Ooooooooooooooffff.
Hmm... *strictly speaking* the examples aren't run, they are just compiled, which I think is still feasible even if you don't have the lib installed, is it?

The main worry I have about that then is that this kinda is doing the statement "This will work on Ubuntu 22.04" which would be a lie if 22.04 is lagging behind over 4 minor versions =/

PhilippMDoerner:
I added a very crude mechanism that I'd copy paste into all other branches that need it and deal with merge conflicts from it as they arise.
Basically I added a check "are you running `nimble examples` on github?" and if so I explicitly ignore certain examples as defined in a seq.
The goal is to extend that seq as needed.
I also added that if you're not on github you automatically get -d:adwminor=4 added so that you don't miss potential breakages in widgets that require higher adwaita versions

can-lehmann:
Looks good! I think the mechanism is fine, the most important thing is that it works :smile:

PR #119, opened 2023-10-21
Add overlay split view

PhilippMDoerner:


PhilippMDoerner:
Note for reference to myself: The examples here aren't compiling because Passwordrow stems from Adwaita 1.4 and Ubuntu 22.04 works with 1.1 at best.
So excluding from the examples for now is necessary

can-lehmann:
Nearly half a year later I finally merged one of these :sweat_smile: Nice work on this PR!

PR #120, opened 2023-10-21
Add passwordrow

PhilippMDoerner:


can-lehmann (approved):

PhilippMDoerner:
Note for reference: The examples here aren't compiling because Passwordrow stems from Adwaita 1.2 and Ubuntu 22.04 works with 1.1 at best.

PR #121, opened 2023-10-21
Add video controls

PhilippMDoerner:


can-lehmann:
What is your opinion on merging the Video and Media Controls examples?

PhilippMDoerner:
Works for me. I was doing most of these in tunnel-vision mode: Get widget, make example according to guidelines, next.

It's a fair assessment to make that the video-controls and video-player example are going to be highly similar.

The main worry I might have is that the video example demonstrates controlling the mediastream on your own as well and I'm not sure whether that might be "too much stuff" in one widget example to have the media-control widget, video widget and buttons to control the media stream in one thing.

If that is still okay then I'll just merge the examples.

Issue #115, opened 2023-10-15
Add the ability to have circular references to Widgets (or at the very least Window)


PhilippMDoerner:
For Features such as CaptureWidget on `SearchEntry` (as discussed [here](https://github.com/can-lehmann/owlkettle/pull/109#issuecomment-1762132276) ) and I think others as well, it is mandatory to be able to have circular references.

Without those implemented, the full featureset of all GTK widgets can not be provided.

I'd like to have an issue about this simply so we can keep this in mind and have somewhere to track possible progress against etc.

Though honestly, how this could possibly work is entirely beyond me.

PhilippMDoerner:
During my waiting times I'm looking into this and I'm wondering whether the first step for this ticket isn't just that we need the ability to query the gui tree we construct via the gui-dsl.

That way you could be able to fetch any Widget from a given gui tree. Imo that'd be step 1, step 2 is be able to check if a given Widget already has a corresponding GtkWidget and fetch that if it does.

Particularly the second step I have absolutely **0** idea how we could achieve that.

can-lehmann:
React solves this issue using "refs": https://react.dev/learn/manipulating-the-dom-with-refs Something similar will work for owlkettle as well.

can-lehmann:
The basic API might look like this:

```nim
let entry = Ref[EntryState]()
gui:
  Window:
    Box:
      Entry:
        reference = entry
      Button:
        proc clicked() =
          entry.moveCursor(0)
          entry.focus()
```

PhilippMDoerner:
I never used react so I'm not fully familiar with the concept. From what I can understand, basically you create a ref first, pass that on to the widget and that "fills" the ref for you?
Given that BaseWidget is the easiest way to make this accessible for most widgets at once, that'd mean add a field "owlRef" or whatever to it and in the beforebuild hook, if hasOwlRef = true, then fill it with... good question. A ref to widgetState? Would make sense I guess since that nets you the entire widget + access to the associated GtkWidget

PhilippMDoerner:
I tried to play around a bit with this (basically passing WidgetState into the Ref and not making it a generic). Question:
Wouldn't a "Ref" approach like this need to sort of work like an observable in rxjs or the like?
By that I mean that you should be able to register callbacks and they fire every time the value of the Ref changes.

[...]

PR #122, opened 2023-10-21
Add toast overlay

PhilippMDoerner:


owlkettle/adw.nim
```diff
@@ -714,7 +717,128 @@ when AdwVersion >= (1, 2) or defined(owlkettleDocs):
   
   export AboutWindow
 
+## Adw.Toast
+proc newToast*(title: string): AdwToast =
+  result = adw_toast_new(title.cstring)
```
can-lehmann, line 722: This should return a wrapper type `Toast` like `MediaStream`/`Pixbuf`/..., to allow for proper garbage collection (especially if the toast is never shown to the user).

PhilippMDoerner: The thing you might run into here is that Toast explicitly hands over ownership to ToastOverlay when it is added.
[Reference](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/method.ToastOverlay.add_toast.html)
> The instance takes ownership of the data, and is responsible for freeing it.

At the same time, Toast is explicitly only to be a helper object for ToastOverlay:
[Reference](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.Toast.html)
> A helper object for [AdwToastOverlay](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.ToastOverlay.html).

So basically there are no scenarios (right now, but honestly I don't think ever) where we should control the memory.
I could add that as a doc comment if you want ?

can-lehmann: > The thing you might run into here is that Toast explicitly hands over ownership to ToastOverlay when it is added.

That should not be an issue, you just need to increment the reference count by one before passing the toast to add_toast.

>  That is unless we assume there can be Toasts that get wasted and not used in a ToastOverlay, so the user creates some and then never hands them to ToastOverlay. Which I'm not sure about if that's a reasonable expectation.

I think especially with the current API, it might actually be a quite common occurence. What if multiple toasts are produced in a single redraw/callback? → depending on the implementation only the last one is shown and the others are discarded.

PhilippMDoerner: This might make things fairly difficult.
A first naive implementation of hooks runs immediately into use-after-free issues where after the toast's dismissal it gets freed by the overlay regardless of the ref count.
If you then update the `toast` variable in the `ToastOverlay`, that leads to the ref to the free'd Toast memory getting removed, thus triggering the destroy hook, which then triggers a use-after-free by trying to unref that reference.

This is pretty hard to overcome, because Toast doesn't know about ToastOverlay's structure and I originally wanted to keep them separate because that's just cleaner and easier to wrap your head around (I mean, this is already not terribly easy).
To avoid this issue, you'd need to `state.toast = nil` the `Toast` reference on `ToastOverlay` when `ToastOverlay` triggers `dismissalCallback` by dismissing the toast.

I am not sure how to solve that.
I've provided the naive implementation for you to look at, but after that I've got no clue.

PhilippMDoerner: I've been chewing on this for a while now, but this might be conceptually more difficult than I expected even going in assuming it'd be not easy.

To avoid a use-after-free, the =destroy hook must know if toast was dismissed by ToastOverlay. Actually, just knowing if it was ever handed over to ToastOverlay would be enough since that means it'll be dismissed and thus free'd eventually, no need for us to take care of that memory.

However, in the =destroy hook we only have access to `ToastObj`, which means `ToastObj` now needs a "wasDismissed" field or so.
So far so good.
Now I need to set `wasDismissed` of a given toast to true in `dismissalCallback` (see the proc `dismissalHandler=`).
`dismissalCallback` however only gets passed the inner `AdwToast` object, not the outer `ToastObj` where the field needs to be set!
I can't conceptually add that field to `data` since `EventObj` doesn't have a field for this.
The only way around this is extending the `EventObj` to contain a ref to `ToastObj`, but I would want to have that be generic in some fashion so that I can reuse that field in other circumstances. Now how that's supposed to work is a tad beyond me =/

can-lehmann: Another approach is just using a regular nim object for `Toast` and the creating the GTK toast in the hook.

```nim
type Toast = ref object
  title: string
  actionName: string
  ...
```

PhilippMDoerner: Good suggestion! I tried it out and it works so far.
With this approach we can save ourselves the hassle of having any hooks for AdwToast in the first place, as it immediately gets handed over to ToastOverlay which then frees it.
Basically removes any memory management from our side, I like it!

docs/widgets_adwaita.md
```diff
@@ -484,6 +484,42 @@ renderable AboutWindow
...
+## ToastOverlay
+
+```nim
+renderable ToastOverlay of BaseWidget
+```
+
+An overlay to display Toast messages that can be dismissed manually and automatically!<br>
+Use `newToast` to create an `Toast`.
+`Toast` has the following properties that can be assigned to:
+- actionName
```
PhilippMDoerner, line 496: On the one hand, we currently don't support that system (can we even at all? I need to jot that down in my notes if we ever will or won't). On the other it's an available property.

Remove? Keep in there?

owlkettle/adw.nim
```diff
@@ -867,6 +870,157 @@ when AdwVersion >= (1, 2) or defined(owlkettleDocs):
...
+    dismissalHandler: dismissalHandler
+  )
+  
+  when AdwVersion >= (1, 2):
+    result.clickedHandler = clickedHandler
+    result.customTitle = customTitle
+  else:
+    let isUsingCustomTitle = not customTitle.isNil()
+    if isUsingCustomTitle:
+      raise newException(LibraryError, "The customTitle field on Toast is not available when compiling for Adwaita versions below 1.2. Compile for Adwaita version 1.2 or higher with -d:adwminor=2 to enable it")
```
PhilippMDoerner, line 908: I employ the strategy here that if you set a value that is only available for certain adwaita versions, you'll get a runtime Exception. Is that alright?

owlkettle/adw.nim
```diff
@@ -867,6 +870,157 @@ when AdwVersion >= (1, 2) or defined(owlkettleDocs):
...
+      
+  when AdwVersion >= (1, 4):
+    result.useMarkup = useMarkup
+  else:
+    let isUsingUseMarkup = useMarkup == true
+    if isUsingUseMarkup:
+      raise newException(LibraryError, "The useMarkup on Toast field is not available when compiling for Adwaita versions below 1.4. Compile for Adwaita version 1.4 or higher with -d:adwminor=4 to enable it")
+
+proc toOwl(adwToast: AdwToast): Toast =
+  when AdwVersion >= (1, 4):
```
PhilippMDoerner, line 922: This is solely necessary for the dismissal-event aka the signal that gets fired when a toast is dismissed.
The user gets to access the Toast that was dismissed (or rather a copy of the data) instead of an AdwToast, therefore the conversion is needed.

Note that I very intentionally here *only* forward the data, I do *not* forward the event handlers themselves because I don't think I can and I don't think we lose anything by not doing so.

owlkettle/adw.nim
```diff
@@ -867,6 +870,157 @@ when AdwVersion >= (1, 2) or defined(owlkettleDocs):
...
+proc toGtk(toast: Toast): AdwToast =
+  result = adw_toast_new(toast.title.cstring)
+  if toast.buttonLabel != "":
+    adw_toast_set_button_label(result, toast.buttonLabel.cstring)
+  adw_toast_set_priority(result, toast.priority)
+  let timeout: cuint = cuint(toast.timeout)
+  adw_toast_set_timeout(result, timeout)
+  
+  # Set Dismissal Handler
+  if not toast.dismissalHandler.isNil():
```
PhilippMDoerner, line 944: I wonder whether I should've moved the contents of those if-blocks into their own procs just to make this a better separation of concerns in terms of the procs

PhilippMDoerner:
Given that this PR is the most advanced one, shall we continue here? 
For the most part I really just need to ask for advice here on how to deal with events thrown from GObject as opposed to from a GTKwidget

can-lehmann:
I think that one of the main problems we need to solve here is how the toasts are passed to the `ToastOverlay` widget. Currently there is the `toasts` field which receives a series of toasts all of which are newly added in the current rerender. Fields should behave in a declarative way though: passing in the same value should lead to no change. I think there are at least two solutions:

- Introduce a new `ToastQueue` type which is passed to the `ToastOverlay`. On every rerender the `ToastOverlay` removes all new toast from the queue and shows them. This also has the advantage of providing an obvious usage pattern for `ToastOverlay` for the user. We already have a similar pattern with e.g. `TextBuffer` or `MediaStream`. 
- Use a ref as proposed in https://github.com/can-lehmann/owlkettle/issues/115 and define an `addToast` method on the ref. This would require implementing refs first though. It also has the issue that toasts could only be added after the initial render.

I currently prefer the `ToastQueue` approach.

can-lehmann:
>  For the most part I really just need to ask for advice here on how to deal with events thrown from GObject as opposed to from a GTKwidget

`TextBuffer` also handles events. Maybe this could be abstracted. https://github.com/can-lehmann/owlkettle/blob/main/owlkettle/widgets.nim#L2317

owlkettle/adw.nim
can-lehmann: I don't think that passing the Toast back to the user when it is dismissed is actually necessary. If you need the dismissed toast, you can just access it via the closure of the event handler, can't you?

PhilippMDoerner:
> > For the most part I really just need to ask for advice here on how to deal with events thrown from GObject as opposed to from a GTKwidget
> 
> `TextBuffer` also handles events. Maybe this could be abstracted. https://github.com/can-lehmann/owlkettle/blob/main/owlkettle/widgets.nim#L2317

Oh I'm an idiot, this wasn't relevant for this PR, this was relevant for https://github.com/can-lehmann/owlkettle/pull/126
There I need it for the eventlisteners on GTKAdjustment. I'll keep that in mind for over there.

owlkettle/adw.nim
PhilippMDoerner: You could make an argument that by passing a Toast you could make accessing the toast more convenient by doing so. But off the cuff, syntactically, yeah you could just do it with a closure.
As in, first define toast, then define dismissal-handler proc which captures the toast object, then assign the dismissalhandler to that toast object.

PhilippMDoerner:
So the remaining points were:
- Move to Toastqueue
- Enable reassignments on dismissalHandler/clickedHandler after initialization, simply because otherwise you can't capture toast in a closure that you assign to those fields.

owlkettle/adw.nim
can-lehmann: I think I prefer just using a closure. Especially since the toast returned by toOwl will be a different object and does not contain the callbacks.

PhilippMDoerner: Alrighty, done

owlkettle/adw.nim
PhilippMDoerner: I did that now for simplification purposes

docs/widgets_adwaita.md
can-lehmann: I think its fine to leave out for now, we can still add it should it become necessary.

owlkettle/adw.nim
```diff
@@ -1286,6 +1289,156 @@ renderable AboutWindow {.since: AdwVersion >= (1, 2).}:
...
+  ## - actionName
+  ## - actionTarget
+  ## - buttonLabel: If set, the Toast will contain a button with this string as its text. If not set, the Toast will not contain a button.
+  ## - detailedActionName
+  ## - priority: Defines the behaviour of the toast. `ToastPriorityNormal` will put the toast at the end of the queue of toasts to display. `ToastPriorityHigh` will display the toast **immediately**, ignoring any others.
+  ## - timeout: The time in seconds after showing the toast after which it is dismissed automatically. Disables automatic dismissal if set to 0. Defaults to 5. 
+  ## - title: The text to display in the toast. Gets hidden if customTitle is set.
+  ## - customTitle: A Widget to display in the toast. Causes title to be hidden if it is set. Only available when compiling for Adwaita version 1.2 or higher.
+  ## - dismissalHandler: An event-handler proc that gets called when this specific toast gets dismissed
+  ## - clickedHandler: An event-handler proc that gets called when the User clicks on the toast's button that appears if `buttonLabel` is defined. Only available when compiling for Adwaita version 1.4 or higher.
```
can-lehmann, line 1415: Should only list the fields that are actually implemented.

docs/widgets_adwaita.md
PhilippMDoerner: Alrighty, removed in that case.

owlkettle/adw.nim
PhilippMDoerner: Removed actionName, actionTarget and detailedActionName as discussed above.

PhilippMDoerner:
Failed because of rate limit, not because this is broken.

owlkettle/adw.nim
```diff
@@ -1286,6 +1289,141 @@ renderable AboutWindow {.since: AdwVersion >= (1, 2).}:
...
+      connectSignal(pointer(result), toast.clickedHandler, "button-clicked")
+
+  when AdwVersion >= (1, 4):
+    adw_toast_set_use_markup(result, toast.useMarkup.cbool)
+
+type ToastQueue* = ref object
+  toasts: seq[Toast]
+  addToOverlay: proc(toast: Toast) {.closure.}
+
+proc isInitialized(queue: ToastQueue): bool = not queue.addToOverlay.isNil()
```
can-lehmann, line 1379: Unused

PhilippMDoerner: Ah, yeah, the entire "addToOverlay" part was from an earlier attempt at getting this to work.
Given it works without that now I summarily removed it

can-lehmann:
Thanks for working on this, I am glad we finally managed to merge it :smile:

PR #125, opened 2023-10-21
Add bin widget (trivial)

PhilippMDoerner:
Honestly, the only reason I wrapped this was because it was there.
I don't see really any practical use in this, but GTK uses it so maybe it'll be useful for inheriting from in the future.

I tested that this works, but the example basically showed absolutely nothing so I didn't add it.


can-lehmann:
In libadwaita, the idea seems to be that other widgets can inherit from it, but that does not work in owlkettle. Also none of the other libadwaita widgets seem to inherit from it, so I don't really see a use of this in owlkettle.

PhilippMDoerner:
Yeah, fair. We can close this. If by some crazy coincidence libadwaita 1.6 starts busting out like 4 different widgets that all inherit from bin we could revisit those to make these 4 bin-classes inherit from this "bin" widget rather than `BaseWidget`, but for now little value. Fair

PR #126, opened 2023-10-22
Add scrollbar widget + Full wrapping of GtkAdjustment

PhilippMDoerner:
This PR's value lies 90% in the fact it wraps GtkAdjustment more fully (or rather: provides a first attempt to), which is useful to more fully encompass functionality of other widgets that make use of it, such as (as per GtkAdjustment docs):

> [GtkSpinButton](https://docs.gtk.org/gtk4/class.SpinButton.html), [GtkViewport](https://docs.gtk.org/gtk4/class.Viewport.html), [GtkScrollbar](https://docs.gtk.org/gtk4/class.Scrollbar.html) and [GtkScale](https://docs.gtk.org/gtk4/class.Scale.html).

HOWEVER: This is mostly the foundation for the PR. 1 thing does not work yet - Event listeners on GtkAdjustment.
That is particularly difficult because unlike with `Toast` in `ToastOverlay` in the other PR, I don't have a GTKWidget as base here.
So the entire existing ecosystem for setting up event listeners is *not* set out for this.

That manifests here in the example compiling, but the event listeners (despite GTK not complaining!) not firing at all.
I'll more this out of the draft stage with updated docs etc. once we can make those event listeners work.


can-lehmann:
I am not sure if we already discussed this at some point, but I do not see the advantage of wrapping `GtkAdjustment`. Mainly because it does not work well with the declarative API. Is there any advantage to wrapping it / where would it be used?

PhilippMDoerner:
The lionshare of the value I see in enabling access to scroll events and being able to know what your current scroll position is.
That knowledge in particular in webdev I needed roughly half a dozen times so far, basically every time you want to add something to the screen but only when the user scrolled fully to the bottom/to the top.

PR #131, opened 2023-10-25
Refactor playground to not use toListFormField + Remove unused imports from examples

PhilippMDoerner:
This takes care of the refactoring we had discussed for playground and for which I had written up a reminder in #106 : https://github.com/can-lehmann/owlkettle/discussions/106#discussioncomment-7350315

Does exactly what the heading states.

This PR should be *mostly* independent from the other 10. I distantly recall one of the other 10 requiring me to implement another toListFormField proc, but I can adjust that.

While I was at it I also noticed a couple examples importing modules they no longer need, so just cleared that up to remove those warning messages.

I also did a slight change to the Scale example, as it really does not look pretty when you add a mark with position top/right to it if it takes up the entire Window. So I put it in a Box() widget with expand: false.

### DIFF
--- a/examples/widgets/picture.nim
+++ b/examples/widgets/picture.nim
@@ -20,7 +20,6 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-import std/[asyncfutures]
 import owlkettle, owlkettle/[playground, adw]
 
 const APP_NAME = "Image Example"
--- a/examples/widgets/picture.nim
+++ b/examples/widgets/picture.nim
@@ -20,6 +20,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+import std/asyncfutures
 import owlkettle, owlkettle/[playground, adw]
 
 const APP_NAME = "Image Example"
--- a/owlkettle.nim
+++ b/owlkettle.nim
@@ -134,42 +134,6 @@ proc remove*(event: EventDescriptor) =
   if g_source_remove(cuint(event)) == 0:
     raise newException(IoError, "Unable to remove " & $event)
 
-proc open*(app: Viewable, widget: Widget): tuple[res: DialogResponse, state: WidgetState] =
-  let
-    state = widget.build()
-    dialogState = state.unwrapRenderable()
-    window = app.unwrapInternalWidget()
-    dialog = state.unwrapInternalWidget()
-  gtk_window_set_transient_for(dialog, window)
-  gtk_window_set_modal(dialog, cbool(bool(true)))
-  gtk_window_present(dialog)
-  
-  proc destroy(dialog: GtkWidget, closed: ptr bool) {.cdecl.} =
-    closed[] = true
-  
-  var closed = false
-  discard g_signal_connect(dialog, "destroy", destroy, closed.addr)
-  
-  if dialogState of DialogState or dialogState of BuiltinDialogState:
-    proc response(dialog: GtkWidget, responseId: cint, res: ptr cint) {.cdecl.} =
-      res[] = responseId
-    
-    var res = low(cint)
-    discard g_signal_connect(dialog, "response", response, res.addr)
-    while res == low(cint):
-      discard g_main_context_iteration(nil.GMainContext, cbool(ord(true)))
-    
-    state.read()
-    if not closed:
-      gtk_window_destroy(dialog)
-    result = (toDialogResponse(res), state)
-  else:
-    while not closed:
-      discard g_main_context_iteration(nil.GMainContext, cbool(ord(true)))
-    
-    state.read()
-    result = (DialogResponse(), state)
-
 proc respond*(state: WidgetState, response: DialogResponse) =
   let
     widget = state.unwrapInternalWidget()
--- a/owlkettle/playground.nim
+++ b/owlkettle/playground.nim
@@ -29,7 +29,7 @@ import ./widgetdef
 import ./widgets
 
 proc toFormField(state: Viewable, field: ptr SomeNumber, fieldName: string): Widget =
-  ## Provides a form field for all number times in SomeNumber
+  ## Provides a form field for all number types in SomeNumber
   return gui:
     ActionRow:
       title = fieldName
@@ -40,11 +40,8 @@ proc toFormField(state: Viewable, field: ptr SomeNumber, fieldName: string): Wid
         proc changed(value: float) =
           field[] = type(field[])(value)
 
-type Range = concept r # Necessary as there is no range typeclass *for parameters*. So `field: ptr range` is not a valid parameter.
-  r is range
-
 proc toFormField(state: Viewable, field: ptr Range, fieldName: string): Widget =
-  ## Provides a form field for any range
+  ## Provides a form field for all range types
   return gui:
     ActionRow:
       title = fieldName
@@ -56,7 +53,7 @@ proc toFormField(state: Viewable, field: ptr Range, fieldName: string): Widget =
           field[] = value
 
 proc toFormField(state: Viewable, field: ptr string, fieldName: string): Widget =
-  ## Provides a form field for string
+  ## Provides a form field for strings
   return gui:
     ActionRow:
       title = fieldName
@@ -66,7 +63,7 @@ proc toFormField(state: Viewable, field: ptr string, fieldName: string): Widget 
           field[] = text
 
 proc toFormField(state: Viewable, field: ptr bool, fieldName: string): Widget =
-  ## Provides a form field for bool
+  ## Provides a form field for booleans
   return gui:
     ActionRow:
       title = fieldName
@@ -76,31 +73,8 @@ proc toFormField(state: Viewable, field: ptr bool, fieldName: string): Widget =
           proc changed(newVal: bool) =
             field[] = newVal
 
-proc toFormField(state: Viewable, field: ptr[auto], fieldName: string): Widget =
-  ## Provides a dummy field as a fallback for any type without a `toFormField`.
-  const typeName: string = $field[].type
-  return gui:
-    ActionRow:
-      title = fieldName
-      Label():
-        text = fmt"Override `toFormField` for '{typeName}'"
-        tooltip = fmt"""
-          The type '{typeName}' must implement a `toFormField` proc:
-          `toFormField(
-            state: Viewable, 
-            field: ptr {typeName}
-            fieldName: string, 
-          ): Widget`
-          state: The <Widget>State
-          field: The field's value to assign to/get the value from
-          fieldName: The name of the field on `state` for which the current form field is being generated
-          
-          Implementing the proc will override this dummy Widget.
-          See the playground module for examples.
-        """
-
 proc toFormField(state: Viewable, field: ptr[enum] , fieldName: string): Widget =
-  ## Provides a form field for enums
+  ## Provides a form field for all enum types
   let options: seq[string] = type(field[]).items.toSeq().mapIt($it)
   return gui:
     ComboRow:
@@ -136,7 +110,7 @@ method view (state: DateDialogState): Widget =
             state.date = date
 
 proc toFormField(state: Viewable, field: ptr DateTime, fieldName: string): Widget =
-  ## Provides a form field for DateTime
+  ## Provides a form field for DateTimes
   return gui:
     ActionRow:
       title = fieldName
@@ -170,7 +144,7 @@ proc toFormField(state: Viewable, field: ptr tuple[x, y: int], fieldName: string
           field[][1] = value.int
 
 proc toFormField(state: Viewable, field: ptr ScaleMark, fieldName: string): Widget =
-  ## Provides a form to display a single entry of type `ScaleMark` in a list of `ScaleMark` entries.
+  ## Provides a form field for the type ScaleMark
   return gui:
     ActionRow:
       title = fieldName
@@ -189,8 +163,7 @@ proc toFormField(state: Viewable, field: ptr ScaleMark, fieldName: string): Widg
           field[].position = enumIndex.ScalePosition
 
 proc toFormField[T](state: Viewable, field: ptr seq[T], fieldName: string): Widget =
-  ## Provides a form field for any field on `state` with a seq type.
-  ## Displays a dummy widget if there is no `toListFormField` implementation for type T.
+  ## Provides a form field for any seq type
   let formFields = collect(newSeq):
     for index, value in field[]:
       toFormField(state, field[][index].addr, fieldName)
--- a/owlkettle/playground.nim
+++ b/owlkettle/playground.nim
@@ -32,7 +32,7 @@ type Range = concept r # Necessary as there is no range typeclass *for parameter
   r is range
 
 proc selfAssign[T](v: var T) = v = v
-proc isObjectVariant(Obj: typedesc[object]): bool =
+proc isObjectVariantType(Obj: typedesc[object]): bool =
   ## Checks if a given typedesc is of an object variant
   ## by checking if you can assign to a field. This is a 
   ## compile-time error when doing this with the fieldPairs iterator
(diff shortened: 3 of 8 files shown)


owlkettle/playground.nim
```diff
@@ -32,71 +32,72 @@ macro getField*(someType: untyped, fieldName: static string): untyped =
   nnkDotExpr.newTree(someType, ident(fieldName))
 
 # Default `toFormField` implementations
-proc toFormField[T: SomeNumber](state: auto, fieldName: static string, typ: typedesc[T]): Widget =
+proc toFormField(state: auto, field: ptr SomeNumber, fieldName: static string): Widget =
```
can-lehmann, line 35: - The `state` parameter is not used anymore.
- `fieldName` does not have to be static

PhilippMDoerner: state is necessary to spawn dialogues, so for some of the overloads (only DateTime).
Thus the parameter is mandatory.

FieldName I fixed though. Also added a delete button to the toFormField for seq's

can-lehmann: Ah, I see. It should be possible to make it of type `Viewable` instead of `auto` though? (so, the compiler does not generate generic overloads for each widget `toFormField` is used in).

PhilippMDoerner: Funnily enough, no, for the same reason as before.
Just tried it out.
Using Viewable *will* blow up the list_view example

can-lehmann: I think the issue here is that `open` has not yet been defined, as it is defined in `owlkettle.nim` which is not imported here. There are two options to fix this: import ../owlkettle (check if its recursive though) or move `open` to another file (e.g. `mainloop`).

PhilippMDoerner: playground.nim can import owlkettle.nim without issue, however that does not appear to be a direct solution.
Even split out into its own file like this:
```nim
import ./widgets
import ./widgetdef
import./bindings/gtk

proc open*(app: Viewable, widget: Widget): tuple[res: DialogResponse, state: WidgetState] =
  let
    state = widget.build()
    dialogState = state.unwrapRenderable()
    window = app.unwrapInternalWidget()
    dialog = state.unwrapInternalWidget()
  gtk_window_set_transient_for(dialog, window)
  gtk_window_set_modal(dialog, cbool(bool(true)))
  gtk_window_present(dialog)
  
  proc destroy(dialog: GtkWidget, closed: ptr bool) {.cdecl.} =
    closed[] = true
  
  var closed = false
  discard g_signal_connect(dialog, "destroy", destroy, closed.addr)
  
  if dialogState of DialogState or dialogState of BuiltinDialogState:
    proc response(dialog: GtkWidget, responseId: cint, res: ptr cint) {.cdecl.} =
      res[] = responseId
    
    var res = low(cint)
    discard g_signal_connect(dialog, "response", response, res.addr)
    while res == low(cint):
      discard g_main_context_iteration(nil.GMainContext, cbool(ord(true)))
    
    state.read()
    if not closed:
      gtk_window_destroy(dialog)
    result = (toDialogResponse(res), state)
  else:
    while not closed:
      discard g_main_context_iteration(nil.GMainContext, cbool(ord(true)))
    
    state.read()
    result = (DialogResponse(), state)
```
Leads to errors
```nim
/home/philipp/dev/owlkettle/owlkettle/playground.nim(140, 10) template/generic instantiation of `gui` from here
/home/philipp/dev/owlkettle/owlkettle/playground.nim(148, 41) Error: type mismatch
Expression: open(state, block:
  let :tmp_2080375130 = DateDialog()
  :tmp_2080375130)
  [1] state: Widget
  [2] block:
  let :tmp_2080375130 = DateDialog()
  :tmp_2080375130: DateDialog

Expected one of (first mismatch at [position]):
[1] proc open(app: Viewable; widget: Widget): tuple[res: DialogResponse,
    state: WidgetState]
[1] proc open(f: var File; filehandle: FileHandle; mode: FileMode = fmRead): bool
[1] proc open(f: var File; filename: string; mode: FileMode = fmRead;
          bufSize: int = -1): bool
[1] proc open(filename: string; mode: FileMode = fmRead; bufSize: int = -1): File
[1] proc open[TMsg](c: var Channel[TMsg]; maxItems: int = 0)
```

can-lehmann: I just tried this and it works fine for me. Not sure why you get type errors.

PhilippMDoerner: Ugh, classic case of me having further changes that lead to this because I was going through multiple different approaches on solving this.
Now this leaves on what to do regarding open().
It's effectively just a proc to open a Dialog in general right?
(So if I were to name it: `openDialog`).

Now the question mostly becomes where to put it.
I don't really have a full understanding of the module split (Note: module comments that declare the "Scope" of a given module would be nice).
Honestly as a helper proc for dialogue I'd have put it in widgets.nim
However that appears to not want to work come hell or high heaven, it keeps finding type mismatches :
```nim
Error: type mismatch
Expression: g_signal_connect(dialog, "destroy", destroy, addr(closed))
  [1] dialog: GtkWidget
  [2] "destroy": string
  [3] destroy: proc (dialog: GtkWidget, closed: ptr bool){.cdecl, noSideEffect, gcsafe.} | proc (ctx: CairoContext){.inline, noSideEffect, gcsafe.} | proc (surface: CairoSurface){.inline, noSideEffect, gcsafe.} | proc (pattern: CairoPattern){.inline, noSideEffect, gcsafe.}
  [4] addr(closed): ptr bool

Expected one of (first mismatch at [position]):
[1] proc g_signal_connect(app: GApplication; signal: cstring; closure, data: pointer): culong
[1] proc g_signal_connect(app: GtkListItemFactory; signal: cstring;
                      closure, data: pointer): culong
[1] proc g_signal_connect(app: GtkSelectionModel; signal: cstring;
                      closure, data: pointer): culong
[1] proc g_signal_connect(widget: GtkEventController; signal: cstring;
                      closure, data: pointer): culong
[3] proc g_signal_connect(widget: GtkWidget; signal: cstring; closure, data: pointer): culong
```
And it makes no sense to me that 3 can not be turned into a ptr when the proc is defined in that module.
Putting it in mainloop seems kinda random, as seems making its own module. It doesn't feel like the kind of place I'd intuitively look for this kind of thing/fits into the way things have been structured so far.

PhilippMDoerner: I fixed it. For some reason having the destroy-proc named "destroy" was something nim did *not* like.
Simple renaming it to `destroyDialog` fixed it.

I moved `open` to widgets.nim since I believe it is most appropriate there as a proc required by a widget to function.
Refactored the entire thing to use Viewable and pushed it just now.

PhilippMDoerner:
Sidenote, 2 other additions I made were:
- I added a delete button at the end of each row for a seq[T] FormField. I thought this would be hard, but turns out I can just put the row Widget of `toFormField` into a `Box` widget and add a delete button after it.
- Apparently range is not covered by current toFormField procs. So I made one that works for all ranges. That required introducing a `Range` concept in playground though, because apparently `proc x(r: range)` is not a valid parameter typeclass to make a generic for all ranges. `proc x[T: range](r: T)` is not workable as that clashes with `proc x(r: auto)`. I am confused and not happy I had to introduce `concept`, but at least the code works :shrug:

PhilippMDoerner:
I had a eureka moment and this did suddenly inspire me to play around again with complete generic form generation for any type.
I succeeded!

We can now fully generate the form for any given type with fields while still having the dummy-widget fallback for all those unknown types e.g. `type Mine = distinct string`. 
This brings with it basically for free support for `Option` types.

See here a generic version of `ScaleMark` where I commented out the custom proc we had before (it also demonstrates it working with Option as the label of a ScaleMark is optional):
Screenshot from 2023-10-26 12-32-16
Screenshot from 2023-10-26 12-32-22
Screenshot from 2023-10-26 12-32-34

The expanding "delete" button on the right is an issue I couldn't quite get rid of and stems from the toFormField of `seq[T]`.
I *think* that's an acceptable sacrifice.

PhilippMDoerner:
And just like that I think I cracked the last piece.
We can now also generate forms for `distinct` types due to the newest `toFormField` proc.

Therefore we can fully auto-generate forms for basically any-type. We're at the point where I'm not sure for what type the dummy-widget with the reminder-text would even be called up anymore.

owlkettle.nim
```diff
@@ -134,42 +134,6 @@ proc remove*(event: EventDescriptor) =
   if g_source_remove(cuint(event)) == 0:
     raise newException(IoError, "Unable to remove " & $event)
 
-proc open*(app: Viewable, widget: Widget): tuple[res: DialogResponse, state: WidgetState] =
```
PhilippMDoerner, line 137: Moved to widgets.nim

owlkettle/widgets.nim
```diff
@@ -3407,6 +3407,43 @@ renderable BuiltinDialog of BaseWidget:
...
+  let
+    state = widget.build()
+    dialogState = state.unwrapRenderable()
+    window = app.unwrapInternalWidget()
+    dialog = state.unwrapInternalWidget()
+  gtk_window_set_transient_for(dialog, window)
+  gtk_window_set_modal(dialog, cbool(bool(true)))
+  gtk_window_present(dialog)
+  
+  proc destroyDialog(dialog: GtkWidget, closed: ptr bool) {.cdecl.} =
```
PhilippMDoerner, line 3420: Renamed because with the name `destroy` it caused compilation errors for reasons I don't understand.

owlkettle/playground.nim
```diff
@@ -20,83 +20,86 @@
...
 
-macro getField*(someType: untyped, fieldName: static string): untyped =
-  nnkDotExpr.newTree(someType, ident(fieldName))
+type Range = concept r # Necessary as there is no range typeclass *for parameters*. So `field: ptr range` is not a valid parameter.
+  r is range
 
-# Default `toFormField` implementations
-proc toFormField[T: SomeNumber](state: auto, fieldName: static string, typ: typedesc[T]): Widget =
-  ## Provides a form field for all number times in SomeNumber
+# Forward Declarations so order of procs below doesn't matter
```
PhilippMDoerner, line 34: Forward Declarations here mostly exist for the procs with  `ptr seq[T]` and `ptr auto`.
I want both of them to have access to *all* other `toFormField` procs, including each other.
This was the only way I could think of that enabled this.

owlkettle/playground.nim
```diff
@@ -123,165 +126,142 @@ method view (state: DateDialogState): Widget =
...
           let (res, dialogState) = state.open(gui(DateDialog()))
           if res.kind == DialogAccept:
-            state.getField(fieldName) = DateDialogState(dialogState).date
+            field[] = DateDialogState(dialogState).date
+
+
+proc toFormField(state: Viewable, field: ptr[distinct], fieldName: string): Widget =
+  ## Provides a form field for any distinct type. 
+  ## The form field provided is the same that the base-type of the distinct type would have.
+  let baseField = field[].distinctBase.addr
```
PhilippMDoerner, line 147: For reference: I double checked that this does not change the actual pointer.
The memory pointed to by baseField and field is the same, just the type that is assumed behind them is different.

For reference, checkout this example:
```nim
import std/typetraits

type X = distinct string

proc debug(x: ptr auto) = 
  echo cast[uint64](x), " - ", x[].type, " - ", x[].distinctBase

let x: X = "lala".X
let y = x.addr
debug y

let z = y[].distinctBase.addr
debug z
```

PhilippMDoerner:
Err maybe just for reference: I do not plan on doing anything else on this unless there's feedback, I'm happy with the current level this is at and can't see anything right now myself that I'd change.

PhilippMDoerner:
Anything else?

examples/widgets/picture.nim
```diff
@@ -20,7 +20,6 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-import std/[asyncfutures]
```
can-lehmann, line 23: This is breaks the example `-d:pixbufAsync`

owlkettle/playground.nim
```diff
@@ -123,165 +126,142 @@ method view (state: DateDialogState): Widget =
...
-      for index, num in state.getField(fieldName):
-        insert(toListFormField(state, fieldName, index, T)){.addRow.}
+      for index, formField in formFields:
+        Box(orient = OrientX) {.addRow.}:
+          insert(formField)
+          Button() {.expand: false.}:
+            icon = "user-trash-symbolic"
+            style = [ButtonDestructive]
+            proc clicked() =
+              field[].delete(index)
```
can-lehmann, line 206: Deleting does not work. GTK outputs `Tried to remove non-child`.

owlkettle/playground.nim
```diff
@@ -123,165 +126,142 @@ method view (state: DateDialogState): Widget =
...
+
+template getIterator*(a: typed): untyped =
+  ## Provides a fieldPairs iterator for both ref-types and value-types
+  when a is ref:
+    a[].fieldPairs
+    
+  else:
+    a.fieldPairs
+    
+proc toFormField(state: Viewable, field: ptr[auto], fieldName: string): Widget =
```
can-lehmann, line 223: This proc errors out for variant types. Instead it should show the dummy field as outlined in the doc comment.

can-lehmann (changes requested):

owlkettle/playground.nim
PhilippMDoerner: That's... new. I literally never encountered this before. However it must've been there for quite a bit because I can't find a commit *now* where that wasn't an issue.

Something super hacky appears in how adding/removing values from `field[]` triggers Widget removals in expander-rows.

PhilippMDoerner: I did a deeper dive with scale.nim and ScaleMark and *none* of this makes sense.
When clicking the delete button, it correctly changes the value in `field[]`.
That correctly triggers the a re-render, which creates a new set of Widgets that get added to `ExpanderRow`, which then update ExpanderRow's `WidgetState` to replace the old GtkWidgets with the new GtkWidgets in `updateAlignedChildren`.

The error occurs when it tries to call the `remove` proc for the Widget created by `toFormField`. Why that suddenly is the case I can not say, none of this makes sense.
My current suggestion is removing the delete button and forcing users to recompile examples in order to remove list-entries.

Or removing auto-form generation again entirely.

owlkettle/playground.nim
PhilippMDoerner: Why not simply introduce a Variant concept so that variant-types get their own `toFormField` proc which returns the dummy FormField widget?
I'll cook up an example in a second

PhilippMDoerner: The most recent comment should've solved it. It introduces an ObjectVariant concept that "catches" all toFormField invocation for Object Variants without their own specific toFormField overloads.

owlkettle/playground.nim
PhilippMDoerner: To document this in a PR comment at the very least:

The problem was that I was using **Box** to add to `ExpanderRow`, not `ActionRow`.

Can.l pointed out this may be related to https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/adw-preferences-group.c#L381

That is because GTK adds a widget with a different proc in adw_preferences_group_add if it's a child of `PreferencesRow` than it does for anything else. Thus you likely also need to remove it in a different manner, which we currently don't do.

owlkettle/playground.nim
```diff
@@ -216,13 +219,14 @@ proc toFormField[T](state: Viewable, field: ptr seq[T], fieldName: string): Widg
...
-            proc clicked() =
-              field[].delete(index) 
+        ActionRow() {.addRow.}:
+            insert(formField) {.addSuffix.}
+            if formField of ActionRow: # GTK allows only removing ActionRow Widgets from ExpanderRow
+              Button() {.expand: false.}:
+                icon = "user-trash-symbolic"
+                style = [ButtonDestructive]
+                proc clicked() =
+                  field[].delete(index)
```
can-lehmann: Indentation

can-lehmann (changes requested):
`ref object` variant types do not work correctly.

can-lehmann (approved):

can-lehmann:
Great work! :smile:

PR #137, opened 2023-11-11
Add hotkeyable ButtonContent component

PhilippMDoerner:
This widget may actually be a bigger deal than I initially thought it was.

Because **it allows you to activate buttons with this as a child via hotkey!**

See the example and docs as for how it does so, the short version is all you need to do is set `useUnderline = true` and surround some piece of text of your choice with a `_` character on each side.
The first letter of that piece of text is used to define a hotkey for that button that is then triggered via `ALT + <That first letter>`.


owlkettle/adw.nim
```diff
@@ -139,6 +139,35 @@ renderable Avatar of BaseWidget:
...
+  hooks useUnderline:
+    property:
+      adw_button_content_set_use_underline(state.internalWidget, state.useUnderline.cbool)
+  
+  hooks canShrink:
+    property:
+      when AdwVersion >= (1, 3):
+        adw_button_content_set_can_shrink(state.internalWidget, state.canShrink.cbool)
+      else:
+        raise newException(ValueError, "Compile for Adwaita version 1.3 or higher with -d:adwMinor=3 to enable the canShrink property for the ButtonContent widget.")
```
can-lehmann, line 169: This does not allow users with adw < (1, 3) to use the widget, because it raises even if canShrink is never set.

owlkettle/adw.nim
```diff
@@ -139,6 +139,35 @@ renderable Avatar of BaseWidget:
       size = 100
       showInitials = true
 
+renderable ButtonContent of BaseWidget:
+  label: string
+  iconName: string
+  useUnderline: bool ## Defines whether you can use `_` on part of the label to make the button accessible via hotkey. If you surround a part of the label text with `_` it will hide the `_` and activate the button if you press ALT + The first key of the surrounded label text. E.g. `_Button_ Text` will trigger the button when pressing `ALT + B`.
```
can-lehmann, line 145: Are you sure that the text needs to be surrounded by `_` rather than needing to insert a single `_` to mark the letter used for the shortcut?

can-lehmann (changes requested):

owlkettle/adw.nim
PhilippMDoerner: Huh, was on autopilot it seems, yeah we never use it that way in property hooks and instead discard.

owlkettle/adw.nim
PhilippMDoerner: I am not, the GTK docs are pretty... meh to understand on that front, at least for me.

Just tried it with only one underscore and that also worked, so appears I was wrong on that one.

examples/widgets/adw/button_content.nim
```diff
@@ -0,0 +1,52 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import owlkettle, owlkettle/[playground, adw]
+
+viewable App:
+  label: string = "_Button_ Text"
```
can-lehmann, line 26:
```suggestion
  label: string = "_Button Text"
```

owlkettle/adw.nim
```diff
@@ -139,6 +139,35 @@ renderable Avatar of BaseWidget:
...
+  hooks useUnderline:
+    property:
+      adw_button_content_set_use_underline(state.internalWidget, state.useUnderline.cbool)
+  
+  hooks canShrink:
+    property:
+      when AdwVersion >= (1, 3):
+        adw_button_content_set_can_shrink(state.internalWidget, state.canShrink.cbool)
+      else:
+        discard
```
can-lehmann, line 169:
```suggestion
```

can-lehmann:
:+1:

PR #146, opened 2023-11-26
Add simple instructions to set up pkg-config for MSYS2

RickBarretto:
Adding simple instructions of how to install the UCRT and CLANG's toolchains for the MSYS2 on Windows.

Thanks to @michaelsbradleyjr.

**Additional links:**
- #144

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -24,7 +24,7 @@ and the `pkg-config` as well.
 
 There are 2 noteworthy toolchain provided by the MSYS2:
 
-A) `CLANG64`: the recommended one for Nim:
+A) `CLANG64`: the recommended one for Nim
 
 ```bash
 $ pacman -Syu mingw-w64-clang-x86_64-toolchain
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -5,7 +5,6 @@ Depending on your operating system, you can install GTK 4 using the following co
 
 - Fedora: `dnf install gtk4-devel libadwaita-devel`
 - Ubuntu: `apt install libgtk-4-dev libadwaita-1-dev`
-- Windows (Msys2): `pacman -S mingw-w64-x86_64-gtk4 mingw-w64-x86_64-libadwaita`
 
 See [the GTK installation guide](https://www.gtk.org/docs/installations/) for more instructions.
 
@@ -15,30 +14,49 @@ We currently recommend installing the development version of owlkettle.
 $ nimble install owlkettle@#head
 ```
 
+## Properly setting up MSYS2 on Windows
+
+For the Windows OS you'll need the MSYS2 
+that you can install from their [official website](https://www.msys2.org/).
+
+Then, you'll need to use a toolchain with essential binaries as compilers, linkers 
+and the `pkg-config` as well.
+
+There are 2 noteworthy toolchain provided by the MSYS2:
+
+A) `CLANG64`: the recommended one for Nim:
+
+```bash
+$ pacman -Syu mingw-w64-clang-x86_64-toolchain
+```
+
+B) `UCRT64`: the recommended one by default for most projects
+
+```bash
+$ pacman -Syu mingw-w64-ucrt-x86_64-toolchain
+```
+
+Then add the GTK-4 and LibAdwaita to your environment.
+
+A) For `CLANG64`
+```bash
+$ pacman -S mingw-w64-clang-x86_64-gtk4 mingw-w64-clang-x86_64-libadwaita
+```
+
+B) For `UCRT64`
+
+```bash
+$ pacman -S mingw-w64-ucrt-x86_64-gtk4 mingw-w64-ucrt-x86_64-libadwaita
+```
+
+Feel free to ask questions about the MSYS2 setup in [How to properly set-up your MSYS2 #144](https://github.com/can-lehmann/owlkettle/discussions/144). 
+
 ## Troubleshooting
 
 1. Ensure that `pkg-config` is installed and the output of the following command includes `-lgtk-4`.
   ```bash
   pkg-config --libs gtk4
   ```
-
-### Properly setting up MSYS2 on Windows
-
-If you want to make sure you have the `pkg-config` in your system, 
-you need to install your MSYS2's environment toolchain.
-
-The 2 noteworthy toolchain provided by the MSYS2 are:
-A) CLANG64: the recommended one for Nim
-B) UCRT64: the recommended one by default for most projects
-
-You can choose one of them by using these commands:
-
-```bash
-$ pacman -Syu mingw-w64-clang-x86_64-toolchain  # CLANG64
-$ pacman -Syu mingw-w64-ucrt-x86_64-toolchain   # UCRT64
-```
-
-Feel free to ask questions about the MSYS2 setup in [How to properly set-up your MSYS2 #144](https://github.com/can-lehmann/owlkettle/discussions/144). 
 
 
 ## Known Incompatibilities
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -5,7 +5,7 @@ Depending on your operating system, you can install GTK 4 using the following co
 
 - Fedora: `dnf install gtk4-devel libadwaita-devel`
 - Ubuntu: `apt install libgtk-4-dev libadwaita-1-dev`
-- Windows (MSYS2): See information bellow
+- Windows (MSYS2): See instructions below
 
 See [the GTK installation guide](https://www.gtk.org/docs/installations/) for more instructions.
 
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -22,6 +22,26 @@ $ nimble install owlkettle@#head
   pkg-config --libs gtk4
   ```
 
+### Properly setting up MSYS2 on Windows
+
+If you want to make sure you have the `pkg-config` in your system,
```
PhilippMDoerner, line 27: `pkg-config` is a package right?

In that case it would be "If you want to make sure that you have the `pkg-config` package in your system, you need to install your MSYS2's environment toolchain.

docs/installation.md
```diff
@@ -22,6 +22,26 @@ $ nimble install owlkettle@#head
   pkg-config --libs gtk4
   ```
 
+### Properly setting up MSYS2 on Windows
+
+If you want to make sure you have the `pkg-config` in your system, 
+you need to install your MSYS2's environment toolchain.
+
+The `UCRT64` is the default environment you should use for any project,
```
PhilippMDoerner, line 30: For continuity with the previous sentence, you may want to mention that these are toolchains. That's beneficial, because then the reader immediately has *proof* in the sentence structure that you mean UCRT64 / CLANG64 when you speak of toolchains. A reader without any related knowledge would have to infer that.

So you could write this something like this:

There are 2 options worth mentioning for your toolchain:
1) The `UCRT64` toolchain, which is the recommended toolchain by default for most projects
2) The `CLANG64` toolchain, which is recommended for Nim

docs/installation.md
```diff
@@ -22,6 +22,26 @@ $ nimble install owlkettle@#head
...
+but `CLANG64` is the recommended one for Nim.
+
+Choose one of them:
+
+```bash
+$ pacman --sync --refresh --sysupgrade mingw-w64-clang-x86_64-toolchain  # CLANG64
+$ pacman --sync --refresh --sysupgrade mingw-w64-ucrt-x86_64-toolchain   # UCRT64
+```
+
+Moreover there is a discussion about this topic made by
```
PhilippMDoerner, line 40: Generally when we have docs such as these we'd want to keep them as short and scannable as possible.
Imagine myself as basically your worst "customer" in that regard as I'm *notorious* for skimming.

Therefore, the authors don't matter nearly as much to me as the information. This could be done in a simple

For further information, check out [this discussion on setting up MSYS2](https://github.com/can-lehmann/owlkettle/discussions/144).

It might make sense to check in with can.l on this. I could see him making the (valid) argument that such a link should never be necessary, as this is already the more detailed installation help and ideally should contain everything you'd need to know.

can-lehmann: Having a link to the discussion for asking questions is fine as long as the main information is contained in the docs. Maybe use "Feel free to ask questions [in this discussion](...)." I agree on not mentioning the authors here. Authorship is recorded via the commits and I additionally try to credit each contributor in the release notes for new versions.

can-lehmann:
I think it makes sense to actually give step by step install instructions for owlkettle on windows. The aim is to have **one** supported way of installing that usually works. If somebody needs help debugging their install it is useful if their setup is as standardized as possible.

E.g. (note that I have never used MSYS2, so this might not be correct)

```markdown
# Installing on Windows using MSYS2

1. Install Msys2 (link to install docs)
2. Install this Msys2 environment: `some command`
3. Install pkg-config: `...`
4. Install GTK: `some command`
5. ...

If you encounter any issues, please check out the troubleshooting steps below. Feel free to ask questions [in this discussion](...).

```

If there are specific troubleshooting steps those should be added to the troubleshooting section.

docs/installation.md
RickBarretto: No, it's not. `pkg-config` is the binary itself, the package are: `mingw-w64-x86_64-toolchain`, `mingw-w64-ucrt-x86_64-pkg-config` and `mingw-w64-clang-x86_64-pkg-config`.

docs/installation.md
RickBarretto: > A reader without any related knowledge would have to infer that.

You are completely right!

docs/installation.md
RickBarretto: Alright, guys! :wink:

docs/installation.md
PhilippMDoerner: In that case my bad, I'd mark this comment chains as resolved if I could :sweat_smile:

RickBarretto: Ok! No problem.

docs/installation.md
RickBarretto: Fixed in: [give better instructions for users that don't have previous knowledge…](https://github.com/can-lehmann/owlkettle/pull/146/commits/08e74c9e6c33a26b45ade520cd94699f37a0b012)

docs/installation.md
RickBarretto: > Generally when we have docs such as these we'd want to keep them as short and scannable as possible.

So, you mean to use the `-Syu` command, instead?

docs/installation.md
PhilippMDoerner: Resolved! I want Bitbuckets resolution interface now to be able to resolve these chains from my end as well!

docs/installation.md
PhilippMDoerner: I wouldn't say this applies to commands. 

These are there for the most part for copy-paste purposes.
Therefore it's fine to keep those as readable as possible and they can be longer - they'll only get read if somebody wants to understand this a bit more in detail.
If they don't folks will just copy-paste-execute-done.

RickBarretto: Fixed in [discussions should only be used for discussions and questions](https://github.com/can-lehmann/owlkettle/pull/146/commits/db05c7a61b5535b6f8edd6c34c9e196733074ce5) and [use a shorter flags for pacman commands](https://github.com/can-lehmann/owlkettle/pull/146/commits/42fadc70de7dd1c54d6ebd276f00e0766bc45fad)

PhilippMDoerner: Alrighty :smile:

RickBarretto:
Now, I have a question @michaelsbradleyjr, in this case, as we're recommending the CLANG64/UCRT64 packages, so I should update the installation of gtk for the respective packages for these environments, right?

RickBarretto:
What about now, guys? @can-lehmann and @PhilippMDoerner.

@michaelsbradleyjr, if you have any suggestion, let me know.

docs/installation.md
```diff
@@ -15,13 +14,51 @@ We currently recommend installing the development version of owlkettle.
...
+
+For the Windows OS you'll need the MSYS2 
+that you can install from their [official website](https://www.msys2.org/).
+
+Then, you'll need to use a toolchain with essential binaries as compilers, linkers 
+and the `pkg-config` as well.
+
+There are 2 noteworthy toolchain provided by the MSYS2:
+
+A) `CLANG64`: the recommended one for Nim:
```
PhilippMDoerner, line 27: Superfluous colon at the end ;-)

Also with the current way of formulating this I get the mild impression that you might be recommending installing both. Afaik you should only install one (?).
Thus it would make sense to rephrase line 25 to make that clearer, like appending "Install one of them".

Rest LGTM!

RickBarretto: Alright!

RickBarretto: Done!

RickBarretto:
What about now, guys?

docs/installation.md
```diff
@@ -15,13 +14,51 @@ We currently recommend installing the development version of owlkettle.
...
 
+## Properly setting up MSYS2 on Windows
+
+For the Windows OS you'll need the MSYS2 
+that you can install from their [official website](https://www.msys2.org/).
+
+Then, you'll need to use a toolchain with essential binaries as compilers, linkers 
+and the `pkg-config` as well.
+
+There are 2 noteworthy toolchain provided by the MSYS2, choose one of them to install:
```
can-lehmann, line 25: You say

> A) `CLANG64`: the recommended one for Nim
> B) `UCRT64`: the recommended one by default for most projects

this is a bit confusing. When would you prefer A over B or otherwise?

docs/installation.md
```diff
@@ -5,7 +5,6 @@ Depending on your operating system, you can install GTK 4 using the following co
 
 - Fedora: `dnf install gtk4-devel libadwaita-devel`
 - Ubuntu: `apt install libgtk-4-dev libadwaita-1-dev`
-- Windows (Msys2): `pacman -S mingw-w64-x86_64-gtk4 mingw-w64-x86_64-libadwaita`
```
can-lehmann, line 8:
```
- Windows (Msys2): See instructions below
```

docs/installation.md
RickBarretto: You're completely right. I'll change this.

I'll remove the `UCRT` so, since the user will be using Nim. And I'll put `UCRT` and `MINGW` as reference if they want to.

I think that delegate the choice for the user would be a bad idea... Instead of that, I'll stick with the `CLANG` choice.

docs/installation.md
RickBarretto: 😉

RickBarretto:
Updated!

michaelsbradleyjr:
Sorry for not replying sooner, caught the flu/whatever and wasn't looking closely at email.

Glad to see this documented, nice work!

One thing that can be considered is adding a recommendation for `pacman -S base-devel git` regardless of those choice of toolchain; note those package names aren't prefixed with e.g. `mingw-w64-clang-x86_64`. That's because they're packages for the base MSYS environment.

https://www.msys2.org/docs/environments/
> The MSYS environment contains the unix-like/cygwin based tools, lives under /usr and is special in that it is always active. All the other environments inherit from the MSYS environment and add various things on top of it.

Most devs working with Nim in CLANG64 (or UCRT64 or ...) will want `git` and the tools included in [`base-devel`](https://packages.msys2.org/package/base-devel?repo=msys&variant=x86_64).

PR #149, opened 2023-12-06
#148 Add custom signal emission utilities

### LINKED ISSUE #148: [Feature Request] Allow setting custom startup/shutdown events in brew procs
## Feature Request

Essentially what I'd love to be able to do would be to have parameters in the brew proc that would allow me to set my own procs to fire when the application starts up (and maybe also when it shuts down).

My explicit usecase would be that I'm currently working on my appster package as you have seen, which sets up a second thread as a "server". I'd love to be able to add a global timeout proc on startup of the application that listens for messages from said server.

Essentially I want to run this piece of code:

```nim
proc addServerListener*[OwlkettleApp: Viewable, SMsg, CMsg](app: OwlkettleApp, data: ServerData[SMsg, CMsg], sleepMs: int = 5) =  
  ## Adds a callback function to the GTK app that checks every 5 ms whether the 
  ## server sent a new message. Triggers a UI update if that is the case.
  proc listener(): bool =
    let msg = data.hub.readServerMsg()
    if msg.isSome():
      app.msg = msg.get()
      discard app.redraw()
    
    const KEEP_LISTENER_ACTIVE = true
    return KEEP_LISTENER_ACTIVE

  discard addGlobalTimeout(sleepMs, listener)
  ```

Right now I'm forced to tell the user "Well you have to know that you must run this in the beforeBuild hook of the App-Widget.
It would be far neater (in my opinion) if I could forego that and just tell them "Add this event to your owlkettle-Startup-Events in the brew proc".

These could also be used for various other things along the lines of setting up resources (e.g. loggers) and destroying them on program shutdown.

## Implementation

Implementation would be pretty trivial. Just add the types:
```nim
type StartUpEvent = proc(app: GApplication, data: ptr AppConfig)
type ShutDownEvent = StartUpEvent
``` 

And then add fields with `seq[StartUpEvent]` and `seq[ShutDownEvent]` in the AppConfig datatype.
Finally add parameters for those fields (default value `@[]`) to all the brew proc and then (depending on the brew proc) execute the events in an appropriate place depending on the brew proc.

PhilippMDoerner:
This is a first draft of adding startup/shutdown events to the various brew procs. I tested them manually, so they should all work.
The approach is as discussed, introduction of a new datatype.
In the GApplication-variants I use the "shutdown" signal, in the others I just execute the events after the runMainloop proc.
In the GApplication-variants I add startup to the "activate" signal, in the others I just execute the events before the runMainloop proc. 

I have not proofread this properly yet, this is just a jump-off point for a basis.
I can definitely confirm that it works at least, as I use this branch with `--path:` to cross-develop it with appster for better integration.

closes #148

### DIFF
--- a/owlkettle/mainloop.nim
+++ b/owlkettle/mainloop.nim
@@ -71,10 +71,11 @@ type
     icons*: seq[string]
     darkTheme*: bool
     stylesheets*: seq[Stylesheet]
-    startupEvents*: seq[ApplicationEvent]
-    shutdownEvents*: seq[ApplicationEvent]
+    startupEvents*: seq[StartUpEvent]
+    shutdownEvents*: seq[ShutDownEvent]
   
-  ApplicationEvent* = proc(widget: Widget, state: WidgetState)
+  StartUpEvent* = proc(widget: Widget, state: WidgetState)
+  ShutDownEvent* = proc(widget: Widget)
   
 # proc registerEvents*(app: GApplication, config: AppConfig, state: WidgetState) =
 #   proc eventCallback(app: GApplication, data: ptr AppData) {.cdecl.} =
@@ -92,9 +93,9 @@ proc execStartupEvents*(config: AppConfig, state: WidgetState) =
   for event in config.startupEvents:
     event(config.widget, state)
     
-proc execShutdownEvents*(config: AppConfig, state: WidgetState) =
+proc execShutdownEvents*(config: AppConfig) =
   for event in config.shutdownEvents:
-    event(config.widget, state)
+    event(config.widget)
 
 proc setupApp*(config: AppConfig): WidgetState =
   if config.darkTheme:
--- a/owlkettle/mainloop.nim
+++ b/owlkettle/mainloop.nim
@@ -78,7 +78,7 @@ type
     startupEvents*: seq[ApplicationEvent]
     shutdownEvents*: seq[ApplicationEvent]
   
-  ApplicationEvent* = proc(state: WidgetState) {.nimcall.}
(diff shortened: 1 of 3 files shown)


owlkettle/mainloop.nim
```diff
@@ -65,11 +65,28 @@ proc loadStylesheet*(path: string, priority: int = DEFAULT_PRIORITY): Stylesheet
...
-  stylesheets*: seq[Stylesheet]
+type 
+  AppConfig* = object of RootObj
+    widget*: Widget
+    icons*: seq[string]
+    darkTheme*: bool
+    stylesheets*: seq[Stylesheet]
+  
+  AppContext* = object
+    config*: AppConfig
```
can-lehmann, line 76: To fix the comment above this would have to be generic over different `AppConfig` types.

can-lehmann: This is weird. It also fails for ref objects:

```nim
type
  A = ref object of RootObj
  B = ref object of A
  
  C[T: A] = object
    x: T

discard C[B](x: B())
```

However when the same pattern is used for the generic instantiation of a proc, it works:

```nim
proc test[T: A](): T =
  result = T()

echo test[A]().repr
echo test[B]().repr
```

Possibly a compiler issue, but even if not, this is inconsistent behaviour.

can-lehmann:
Is this ready for merging from your side?

PhilippMDoerner:
I honestly expected more to be wrong with it and require a new approach with this more being like a first stab at the problem to learn looking for more solutions. Particularly because I wrote most of this code while sick.

If this woks I'm fine with it, I still don't have the mental capacity to properly proof-read my work here but I'm happy to rely on your review of it :sweat_smile:

PR #157, opened 2024-03-01
extend AboutWindow

nylki:
This PR implement previously missing bindings for the following fields in `AboutWindow` (Adwaita):

- applicationIcon
- releaseNotes
- comments
- debugInfo
- developers
- designers
- artists
- documenters
- licenseType
- legalSections
- credits
- acknowledgements
- links

I also added an example for the Adwaita AboutWindow with a screenshot and everything. :slightly_smiling_face: 
Originally, I planned to implement all the missing fields.
But since I lack experience with both nim and GTK development, I didn't feel comfortable touching the `translators` field because it appears to ties into app localization which I was not sure if it was even supported in owlkettle yet.

Some other things I wasn't sure about when implementing the bindings:

-  `legalSections`: I defined a custom type to be easier able to fill the list. However I am not sure if thats the way to go here, or if it could be made more concise with standard dictionary/tuple types (which I didn't manage to do).

- optional section name in `credits`: I didn't make it optional in this MR, as I wasn't sure what's the best way to deal with optional values is in nim/owlkettle (see above).


docs/widgets.md
```diff
@@ -948,6 +948,10 @@ A text editor with support for formatted text.
 - `acceptsTab: bool = true`
 - `indent: int = 0`
 
+###### Events
+
+- changed: `proc ()`
```
nylki, line 953: Not sure why this section was added after running `nimble genDocs`

owlkettle/adw.nim
```diff
@@ -818,8 +849,20 @@ when AdwVersion >= (1, 2) or defined(owlkettleDocs):
     issueUrl: string
     website: string
     copyright: string
-    license: string
-    
+    license: string ## Sets the license as a custom text if it can’t be set via licenseType (when set licenseType will be set to GTK_LICENSE_CUSTOM).
+    licenseType: LicenseType ## Sets the license for from a list of known license.
+    legalSections: seq[LegalSection] ## Adds an extra section to the Legal page. Extra sections will be displayed below the application’s own information. The parameters copyright, licenseType and license will be used to present the it the same way as AboutWindow:copyright, AboutWindow:licenseType and AboutWindow:license are for the application’s own information. See those properties for more details. This can be useful to attribute the application dependencies or data.
```
can-lehmann, line 854: Using the GTK doc comments is an issue as 

1. They often do not apply to owlkettle. E.g. in this case copyright, licenseType and license are not parameters to legalSections, as it is a field, not a function. 
2. GTK is LGPL licensed. I am not a lawyer, but from my understanding, as this license also applies to the (doc) comments, using GTK's doc comments would require owlkettle to be relicensed under LGPL (or a compatible license).

examples/widgets/adw/about_window.nim
```diff
@@ -0,0 +1,108 @@
...
+                version = "1.0.0"
+                applicationIcon = "application-x-executable"
+                supportUrl = "https://github.com/can-lehmann/owlkettle/discussions"
+                issueUrl = "https://github.com/can-lehmann/owlkettle/issues"
+                website = "https://can-lehmann.github.io/owlkettle/README"
+                links = @{
+                  "Tutorial": "https://can-lehmann.github.io/owlkettle/docs/tutorial.html",
+                  "Installation": "https://can-lehmann.github.io/owlkettle/docs/installation.html"
+                }
+                comments = """My Application demonstrates the use of the Adwaita AboutWindow. Comments will be shown on the Details page, above links. <i>Unlike</i> GtkAboutDialog comments, this string can be long and detailed.It can also contain <a href='https://docs.gtk.org/Pango/pango_markup.html'>links</a> and <b>Pango markup</b>."""
```
can-lehmann, line 53: Same issue as with the doc comments

examples/widgets/adw/about_window.nim
```diff
@@ -0,0 +1,108 @@
...
+import std/options
+
+viewable App:
+  discard
+
+method view(app: AppState): Widget =
+  result = gui:
+    Window:
+      title = "About Dialog Example"
+      defaultSize = (400, 200)
```
can-lehmann, line 33: I would consider increasing the window size a bit, as the small size seems to cause weird behavior when attempting to move the about window.

can-lehmann (changes requested):
Looks pretty good already :smile: I played with the example a bit and it works quite well. Here are some small comments.

owlkettle/adw.nim
```diff
@@ -818,8 +849,20 @@ when AdwVersion >= (1, 2) or defined(owlkettleDocs):
...
+    releaseNotes: string
+    comments: string
+    debugInfo: string
+    developers: seq[string]
+    designers: seq[string]
+    artists: seq[string]
+    documenters: seq[string]
+    credits: seq[(string, seq[string])] ## Adds additional credit sections with customizable titles
+    acknowledgements: seq[(string, seq[string])] ## Adds acknowledgment sections with customizable titles
+    links: seq[(string, string)] ## Adds additionals links placed in the details section
```
can-lehmann, line 865: I would use tuples with named fields (`tuple[x: float]`) instead of anonymous tuples. Otherwise the purpose of the fields is not immediately obvious.

can-lehmann:
Regarding the `legalSections` field: Using a custom type a completely valid approach imo.

Regarding optional values: I would say that if possible, using the empty string value as the NULL value. If the widget has different behavior when passing an empty string vs. passing NULL, use std/options.

nylki:
I appreciate you taking the time to look into the PR! :)
I added a couple commits addressing all your comments.

> Regarding optional values: I would say that if possible, use the empty string value as the NULL value. If the widget has different behavior when passing an empty string vs. passing NULL, use std/options.

Thanks! As it turns it in the case of the credits section ( `adw_about_window_add_credit_section`), using an empty string ("") for the title, already works as intended.

examples/widgets/adw/about_window.nim
nylki: I did increase the size as you requested, but did not observe any weird behaviour myself (Fedora 39, libadwaita 1.4.3-2).

owlkettle/adw.nim
nylki: I adjusted the comments to not be a direct copy of the GTK docs and to better fit owlkettle usage in some cases.

owlkettle/adw.nim
```diff
@@ -809,6 +809,37 @@ renderable StatusPage of BaseWidget:
...
+  LICENSE_GPL_2_0_ONLY = 9
+  LICENSE_GPL_3_0_ONLY = 10
+  LICENSE_LGPL_2_1_ONLY = 11
+  LICENSE_LGPL_3_0_ONLY = 12
+  LICENSE_AGPL_3_0 = 13
+  LICENSE_AGPL_3_0_ONLY = 14
+  LICENSE_BSD_3 = 15
+  LICENSE_APACHE_2_0 = 16
+  LICENSE_MPL_2_0 = 17
+  LICENSE_0BSD = 18
```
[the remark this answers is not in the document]
nylki: Converted them to PascalCase, however I am not sure how well it works for licenses like `LicenseLGPL2_1_Only` :thinking:.
let me know if they are fine now or if can think of a better way for the ones containing both appreviations and decimal numbers.

can-lehmann:
Looks good, thanks for contributing!

I fixed some typos before merging. I also changed the example values for the `license` field from `LicenseArtistic` to `LicenseMIT_X11`, as I am not familiar with the Artistic license and was not sure if it is compatible with MIT (Owlkettle) / LGPL (GTK) (it might very well be, I just don't know). So just to be sure and not accidentally imply a compatibility to potential users, I changed it to `LicenseMIT_X11`.

PR #160, opened 2024-04-10
Added overshot and reached signals to scrolled window

beef331:
Also added a `config.nims` at the root to make it easier to test single files.


beef331:
Is there a reason the `.gitignore` does not just ignore all extensionless files, it would make it much cleaner?

can-lehmann:
Works quite nicely. I made some small improvements before merging.

> Is there a reason the .gitignore does not just ignore all extensionless files, it would make it much cleaner?

Not really, it just never occured to me that one could do that. If you want to, you can open a PR.

PR #166, opened 2024-06-17
#115 WIP Request for Code-Style Review when it comes to DSL

PhilippMDoerner:
This PR is a draft to demonstrate how I plan on acting while in the DSL part of the code.

Given that macros and DSLs in general have a tendency to be pretty mean in terms of complexity, I developed a few habits over the years to structure those well enough so I can more easily get back into it after a while.

I plan on introducing some of those in the owlkettle DSL because I believe them to be beneficial to the overall project and for readability.

I wanted to open up this PR to already show you how I tend to act on every proc I touch when it comes to these to give you the chance to veto any of these habits up front.

In general:
-Add doc comments to any proc involved in code generation. Even private procs.
Procs that generate NimNodes and thus Code directly I also annotate with a pattern that demonstrates *what* they generate (see doc comment in the code)
- A variable should always only hold one type of NimNodeKind.
I just found it a ton easier to reason about what a given NimNode is if I can know for sure what NimNodeKind it is.
That's visible here in how I replaced the bits of result with var fieldName until the final code generation step
- Separate sections in a proc into either smaller procs with fitting names, or at the very least split with doc comments. Code-generating related procs tend to become hard to see what they're about really fast, which is why I have this rule so you have *something* that allows you to see at a glance what this codeblock is about.

### DIFF
--- a/owlkettle/widgetdef.nim
+++ b/owlkettle/widgetdef.nim
@@ -22,7 +22,7 @@
 
 # Macros used to define new widgets
 
-import std/[macros, strutils, tables]
+import std/[macros, strutils, tables, sets, options]
 when defined(nimPreviewSlimSystem):
   import std/assertions
 import common
@@ -81,6 +81,45 @@ proc redraw*(viewable: Viewable): bool =
   if not newWidget.isNil:
     viewable.viewed = newWidget
     result = true
+
+type
+  StateRef*[T] = ref object
+    state: T
+    observers: HashSet[proc(state: T)]
+
+proc hasRef*[T](stateRef: StateRef[T]): bool = not stateRef.state.isNil()
+
+proc newRef*[T](): StateRef[T] = 
+  StateRef[T](observers: initHashSet[proc(state: T)]())
+
+proc unwrap*[T](stateRef: StateRef[T]): T =
+  stateRef.state
+
+proc setRef*[T](stateRef: StateRef[T], state: T) =
+  if stateRef.isNil():
+    return
+  
+  let isStateChange = stateRef.state == state
+  if not isStateChange:
+    return
+  
+  stateRef.state = state
+  for observer in stateRef.observers:
+    observer(stateRef.state)
+
+proc widget*[T](stateRef: StateRef[T]): Option[GtkWidget] = 
+  echo "T is: ", $T
+  when T is Renderable:
+    some stateRef.state.internalWidget
+  else:
+    none(GtkWidget)
+
+proc subscribe*[T](stateRef: StateRef[T], observer: proc(state: T)) =
+  stateRef.observers.incl(observer)
+
+proc unsubscribe*[T](stateRef: StateRef[T], observer: proc(state: T)) =
+  stateRef.observers.excl(observer)
+
 
 type
   WidgetKind = enum WidgetRenderable, WidgetViewable
@@ -147,6 +186,10 @@ proc value(field: Field): NimNode =
   result.copyLineInfo(field.lineInfo)
 
 proc stateName(def: WidgetDef): string = def.name & "State"
+proc camelCase(s: string): string =
+  result = s
+  result[0] = ($result[0]).toLower()[0]
+proc refName(def: WidgetDef): string = def.name.camelCase() & "Ref"
 
 proc widgetBase(def: WidgetDef): NimNode =
   if def.base.len == 0:
--- a/owlkettle/widgetdef.nim
+++ b/owlkettle/widgetdef.nim
@@ -442,18 +442,36 @@ proc substituteWidgets(node: NimNode): NimNode =
         result.add(substituteWidgets(child))
 
 proc genState(def: WidgetDef): NimNode =
-  result = newTree(nnkRecList)
+  ## Generates the WidgetState for the widget defined in def.
+  ## The type is generated according to the pattern:
+  ## ```
+  ## type <WidgetName>State = ref object of X
+  ##   <WidgetName>Ref: StateRef[<WidgetName>State]
+  ##   --- Repeat per field in def.fields - start ---
+  ##     <fieldName>: <fieldType>
+  ##   --- Repeat per field in def.fields - end ---
+  ##   --- Repeat per event in def.events - start ---
+  ##     <eventName>: Event[proc(<eventParameters>)]
+  ##   --- Repeat per event in def.events - end ---
+  ## ```
+  var fieldNodes = newTree(nnkRecList)
+  
+  # Add data fields  
   for field in def.fields:
     var fieldType = field.typ
     if def.kind == WidgetRenderable:
       fieldType = substituteWidgets(fieldType)
-    result.add(newTree(nnkIdentDefs, [
+    fieldNodes.add(newTree(nnkIdentDefs, [
       ident(field.name).newExport(not field.isPrivate),
       fieldType,
       newEmptyNode()
     ]))
+    
+  # Add event fields
   for event in def.events:
-    result.add(event.genIdentDefs())
+    fieldNodes.add(event.genIdentDefs())
+  
+  # Generate Type
   result = newTree(nnkTypeDef, [
     ident(def.stateName),
     newEmptyNode(),
@@ -461,7 +479,7 @@ proc genState(def: WidgetDef): NimNode =
       newTree(nnkObjectTy, [
         newEmptyNode(),
         newTree(nnkOfInherit, def.stateBase),
-        result
+        fieldNodes
       ])
     ])
   ])


PhilippMDoerner:
The last 2 commits implement an example on how to use the stateRef by implementing a feature we previously didn't have - keyCaptureWidget on search.

It allows "forwarding" keypress-events that happen while another widget is in focus to the SearchEntry Widget.
This allows in the example to have the list of items underneath the searchentry in focus, and while typing with it in focus the searchEntry fills with text.

From what I can tell it works correctly.

Some thoughts where I'd like to have your opinion:
- 1. As written in discord, StateRef is **the only** field where assignments can have 2 meanings: Assigning to a stateRef field means "Fill the StateRef instance I passed into the widget with a reference". Assigning to anywhere else means "Use the StateRef instance I passed into the widget for whatever". Unless you *know* stateRef is the "source" for the Ref and the other fields are the "sinks" for their usage, this is very non-obvious
- 2. To properly deal with subscribing/unsubscribing as stateRef changes on a widget, I need to do things in *both* the build *and* the update hook of stateRef. In the build-hook I need to do the initial subscription with observer. In the update-hook I need to first check if the stateRef changed. If it did, I want to unsubscribe the observer from the old stateRef and then add the observer to the new stateRef. That's a bit wordy and I'm not 100% if that's accurate.

PhilippMDoerner:
Just [had a chat with leorize](https://discord.com/channels/371759389889003530/371759389889003532/1253757624567074866): He suggests using the `as` operator for the semantics on assigning a ref. And I very much agree that the semantics look a lot clearer.

This would basically change this:
```nim
      ScrolledWindow:
        stateRef = app.keyCaptureRef
        ... other stuff....
``` 
to this
```nim
      ScrolledWindow as app.keyCaptureRef:
        ... other stuff....
```

It'd require some finageling with the DSL I think but in my eyes this improves clarity on where a ref comes from quite a bit.

PhilippMDoerner:
I did some experimenting about how to implement this. 
With the newest commit (which can easily be reverted) we can now have the `as` Syntax, see search_entry.nim in this PR:
```nim
      ScrolledWindow as app.keyCaptureRef:
        ListBox:
          ... other stuff ...
```

I wrote an in-depth explanation of what I did and the reasoning behind it in the commit message if anything further is necessary.

can-lehmann:
> ```nim
>  ScrolledWindow as app.keyCaptureRef:
>    ... other stuff....
> ```

This looks pretty nice! How does it interact with adders and the `insert` statement?

PhilippMDoerner:
The way it currently is implemented I **think** it doesn't interact with the insert statement at all.
By which I mean you likely can't use this syntax with insert at the moment, that'd require additional changes to what I did in `parseGui`. 

As for with adders, in which sense do you mean?

owlkettle/widgetdef.nim
```diff
@@ -31,7 +31,9 @@ import bindings/gtk
 type
   Widget* = ref object of RootObj
     app*: Viewable
-  
+    hasStateRef*: bool
+    valStateRef*: StateRef
```
can-lehmann, line 35: If you use custom syntax, you can just have a single `stateRef: StateRef` field here.

PhilippMDoerner: Given that stateRef is very much an optional field and a ref-type which might be nil, mind if I use `Option[StateRef]`?

can-lehmann:
> As for with adders, in which sense do you mean?

My question is if using `as` breaks the adder syntax. Where is the adder specified when using `as`? `Widget {. ... .} as ...` or `Widget as .... {. ... .}`?

can-lehmann:
> I wanted to open up this PR to already show you how I tend to act on every proc I touch when it comes to these to give you the chance to veto any of these habits up front.

I think that such refactoring should be made in a separate PR as they are not really related to adding Refs (or given the current title of this PR, adding refs should be a separate PR).

owlkettle/widgetdef.nim
can-lehmann: `nil` should already express the absence of a `StateRef`.

owlkettle/guidsl.nim
```diff
@@ -93,6 +94,28 @@ proc parseGui(node: NimNode): Node =
         result = node[0].parseGui()
       for it in 1..<node.len:
         result.children.add(node[it].parseGui())
+    of nnkInfix: # For expressions like "<Widget> as <stateRefVariable>"
+      let asNode = node[0]
+      let isRefAssignmentExpression = asNode.kind == nnkIdent and $asNode == "as"
```
can-lehmann, line 99: should use eqIdent

owlkettle/guidsl.nim
```diff
@@ -45,6 +45,7 @@ type
       of NodeWidget:
         widget: seq[string]
         adder: Adder
+        widgetRefVar: Option[NimNode] # Contains a variable of type StateRef that may receive a reference to the constructed widget.
```
can-lehmann, line 48:
```suggestion
        stateRef: NimNode
```

And all changes needed to make this compile. `stateRef.isNil` should be used instead of optionals.

examples/widgets/search_entry.nim
```diff
@@ -73,7 +74,7 @@ method view(app: AppState): Widget =
             app.text = ""
             app.filteredItems = app.items
           
-      ScrolledWindow:
+      ScrolledWindow as app.keyCaptureRef:
```
PhilippMDoerner, line 77: Turns out this will not fill the keyCaptureRef if you use `ScrolledWindow() as app.keyCaptureRef`. Which must be fixed.

PhilippMDoerner:
> My question is if using as breaks the adder syntax. Where is the adder specified when using as? Widget {. ... .} as ... or Widget as .... {. ... .}?

Right now this breaks the DSL, should be fixable by adjusting some of the parsing again.

PhilippMDoerner:
This is going to interact with adder or rather pragmas in general in a really ugly way that I haven't yet found a simple way to solve.

Parsing of pragmas happens in `    of nnkPragmaExpr:` in line 123.
However, adding the possibility of an infix approach, that means expanding that section because you might be parsing a pragma **either** as part of a "normal" line such as `<Widget> {.addRow.}` or as part of an infix line such as `<Widget> as <someRef> {.addRow.}`.

PhilippMDoerner:
@can-lehmann Fixed the interaction with adders not working, example for that is added to SearchEntry. I likely shouldn't put that there but inspiration didn't hit me on how/where else to demonstrate it.

can-lehmann:
> I likely shouldn't put that there but inspiration didn't hit me on how/where else to demonstrate it.

I recently wrote a guide on the GUI DSL. You need to add a short subsection on state refs there anyways, so maybe add it there. Of course add a simple example first and then say "To get the reference of a widget with adders, the following syntax is used: " <example>

can-lehmann:
Another issue you need to solve is what happens when a widget subsribes to a state ref but is deleted afterwards. Currently the observer still remains registered causing a use after free as far as I can see.

PhilippMDoerner:
> Another issue you need to solve is what happens when a widget subsribes to a state ref but is deleted afterwards. Currently the observer still remains registered causing a use after free as far as I can see.

The obvious and clean way to solve that efficiently architecturally speaking is having a destroy hook.
Now how that's supposed to work I don't even have a concept for.

owlkettle/widgetdef.nim
```diff
@@ -82,6 +87,46 @@ proc redraw*(viewable: Viewable): bool =
     viewable.viewed = newWidget
     result = true
 
+proc hasRef*(stateRef: StateRef): bool = not stateRef.isNil() and not stateRef.state.isNil()
+
+proc newRef*(subscribers: varargs[proc(state: WidgetState) {.closure.}]): StateRef =
```
can-lehmann, line 92:
```suggestion
proc newStateRef*(subscribers: varargs[proc(state: WidgetState) {.closure.}]): StateRef = 
```

owlkettle/widgetdef.nim
```diff
@@ -82,6 +87,46 @@ proc redraw*(viewable: Viewable): bool =
...
+
+proc unwrapInternalWidget*(stateRef: StateRef): GtkWidget =
+  if stateRef.isNil() or stateRef.state.isNil():
+    return nil.GtkWidget
+  
+  return stateRef.state.unwrapInternalWidget()
+
+proc setRef*(stateRef: StateRef, state: WidgetState) =
+  if stateRef.isNil():
+    return
```
can-lehmann, line 106: I don't see why you check arguments for nil everywhere. At least from a code style perspective this seems like a very uncommon choice in Nim.

PhilippMDoerner: A nil check is always preferable over a segfault. The reason why I typically go for optionals when a specific instance actually is allowed to not exist (which is the case for stateRef instances, a widget may not have one) is to make handling of it explicit.

With a ref-type you never know. It blew up in my face enough times during development that just having the nil-checks there was the far safer option.

owlkettle/guidsl.nim
```diff
@@ -93,6 +94,35 @@ proc parseGui(node: NimNode): Node =
...
+      let widgetName = case node[1].kind:
+        of nnkIdent: node[1]
+        of nnkCall: node[1][0]
+        else: 
+          error("Tried to use 'as' with invalid syntax", node)
+          newEmptyNode() # Forces the compiler to acknowlege that all branches of the case statement return a NimNode
+      
+      let stateRefVar = case node[2].kind:
+        of nnkPragmaExpr: node[2][0]
+        else: node[2]
```
can-lehmann, line 110: This just ignores the adder, doesn't it?

If you want to use the syntax `Widget {.adder.} as ...`. Please report an error if the user tries to use `Widget as ... {.adder.}`.

owlkettle/guidsl.nim
```diff
@@ -220,6 +250,14 @@ proc gen(node: Node, stmts, parent: NimNode) =
...
       for child in node.children:
         child.gen(body, name)
+      
+      let hasStateRef = not node.stateRef.isNil()
+      if hasStateRef:
+        let refVar = node.stateRef
+        let refAssignment = quote do:
+          `name`.stateRef = `refVar`
+        body.add(refAssignment)
+
```
can-lehmann, line 260:
```suggestion
      if not node.stateRef.isNil:
        body.add: quote do:
          `name`.stateRef = `node.stateRef`
      
```

complexity left over from when you were using optionals.

owlkettle/widgetdef.nim
can-lehmann: The caller should check. Having the callee just silently do nothing is probably worse than a segfault, because if the stateRef is nil, that is a logic error.

PhilippMDoerner: I'll stick to it as it's the code-style in the rest of the project, but generally I'd say I highly prefer Options due to their explicit nature of informing you "this might not exist", while in all other cases you just have to know that sth might not exist.

owlkettle/guidsl.nim
PhilippMDoerner: It does not in fact. Like, I didn't think deeper about that fact, but the DSL combination that gets created by `Widget {.somepragma.}  as ...` appears to run into the else block in line 105.
```nim
        else: 
          error("Tried to use 'as' with invalid syntax", node)
```

At least when I added
```nim          
          ListBoxRow() {.addRow.} as app.dummyRef:
            Label(text = "Static last")
```
To the search_entry example it blew up in my face with exactly that error message.

owlkettle/guidsl.nim
PhilippMDoerner: >           `name`.stateRef = `node.stateRef`
You can't do this replacement, this won't compile due to
> /home/isofruit/dev/owlkettle/owlkettle/guidsl.nim(144, 20) Error: type mismatch: got 'array[0..-1, empty]' for '[]' but expected 'seq[Node]'

I assume this is quote do being inconsistent in its behavior if you don't have an explicit var that you want to insert.

PhilippMDoerner:
Any thoughts on how a destroy-hook in owlkettle would be implemented?

can-lehmann:
You might need to attach it to the gtk widget directly (via destroy notify I believe), as the lifetime of the `WidgetState` is not necessarily equal to the lifetime of the gtk widget.

PhilippMDoerner:
Memo to me for later in another PR:
I'll need to see if I can create some procs that make performing this setup easier when I implement a second feature that makes use of this.

can-lehmann:
> You might need to attach it to the gtk widget directly (via destroy notify I believe), as the lifetime of the `WidgetState` is not necessarily equal to the lifetime of the gtk widget.

The other option would be making sure that the lifetime of the gtk widget is always at least equal (or longer) than the lifetime of the WidgetState (by using g_object_ref/unref correctly). This might actually be a nice thing to do. Then you could use `=destroy`.

PhilippMDoerner:
>  The other option would be making sure that the lifetime of the gtk widget is always at least equal (or longer) than the lifetime of the WidgetState (by using g_object_ref/unref correctly). This might actually be a nice thing to do. Then you could use =destroy.

I'll claim incompetence on that one, this goes over my head on how I'd even make sure that is correct. It also conceptually makes no sense to me because in my head WidgetState persists across various iterations of Widgets, it just gets updated. That's one of the reason why the usage of StateRef requires you assigning it to WidgetState.

PhilippMDoerner:
Okay, so I found it rather easy to add destroy-hooks to be parseable, just add to the HookEnum and to parseHookKind and you're done.

I'm getting the feeling doing this would be infinitely easier for you than for me because I'm massively struggling even vaguely getting an idea on where to add things and the intended place for additions.

Issue #167, opened 2024-07-22
MacOS cannot find libadwaita


mrunion:
I am trying a simple window and trying to use Adwaita. The error I get when running "nim c..." is:

```
ld: library 'adwaita-1' not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I have GTK4 and libAdwaita installed via brew. The code for the app is simple:

```nim
import owlkettle
import owlkettle/adw

viewable App:
    text: string

method view(app: AppState): Widget =
    result = gui:
        Window:
            title = "Simple Window"

adw.brew(gui(App()))
```

If I use GTK4 itself by removing the "adw.brew..." and just using "brew...", and removing the adw import the code runs fine.

pkg-build shows the following output when running "pkg-config --libs libadwaita-1"

```
-L/opt/homebrew/Cellar/libadwaita/1.5.2/lib -L/opt/homebrew/Cellar/gtk4/4.14.4/lib 
-L/opt/homebrew/Cellar/pango/1.54.0/lib -L/opt/homebrew/Cellar/harfbuzz/9.0.0/lib 
-L/opt/homebrew/Cellar/gdk-pixbuf/2.42.12/lib -L/opt/homebrew/Cellar/cairo/1.18.0/lib 
-L/opt/homebrew/Cellar/graphene/1.10.8/lib -L/opt/homebrew/Cellar/glib/2.80.4/lib 
-L/opt/homebrew/opt/gettext/lib 
-ladwaita-1 -lgtk-4 -lpangocairo-1.0 -lpango-1.0 -lharfbuzz -lgdk_pixbuf-2.0 
-lcairo-gobject -lcairo -lgraphene-1.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0 -lintl
```

Am I doing something wrong or is MacOS not supporting Adwaita, or any other clues?

Thanks in advance!

can-lehmann:
It seems like the shared library file is missing. From the pkg-config output I would assume that it is supposed to be in `/opt/homebrew/Cellar/libadwaita/1.5.2/lib`. Does this folder contain the shared library? Can you compile & link libadwaita applications written in C using gcc?

mrunion:
The dynlib is in that location, so I am assuming this is correct? (I am on Apple Silicon, so not sure if that's relevant or not here.)

Contents of /opt/homebrew/Cellar/libadwaita/1.5.2/lib
```
[...]

PR #168, opened 2024-07-24
Fixed compilation of Adwaita-based GUI

mrunion:
Fixed the issue of the search path not including the libadwaita path when compiling. The issue was reported and discussed in #167.


can-lehmann:
Ah, that explains it. Thanks for the fix!

